PR #7, opened 2018-09-28
Allow Select to quit without selecting an Item with Esc or q

bn3t:
Hi,
I want to use Select but allow the user to quit without selecting an Item. I have added interact_opt() and interact_on_opt() methods to Select. These methods return Result<Option<usize>> instead of Result<usize>.


bn3t:
Please consider my pull request. At least comment.

nicohman:
Merged! Thanks for the contribution.

PR #13, opened 2019-02-24
Fixes an issue with select not clearing enough lines if a selectable …

twick00:
…option spans more than a single line.


twick00:
First open-source contribution so go easy.

PR #32, opened 2019-09-20
Added FuzzySelection type interaction, updated selection prompts to have multiline prompts, added DateTimeSelector

slaterb1:
This interaction is similar to `Selection` but allows the user to type in characters to match against to filter to selections. For simplicity, the selected option is returned instead of it's position (that is relative depending on the number of items in the reduced set).

Moving the cursor through the selections is done via the arrow keys (same as Selection), but vim bindings (hjkl) now only effect filtering the words. 

### Added further enhancements on my fork
Now `Select`, `Checkboxes`, and `FuzzySelect` can display selections that span multiple lines and have a custom set offset.

`FuzzySelect` now includes ability to use casing in matches.

### Added DateTimeSelector on my fork
This is a new interactions that allows you to select a date by updating a displayed datetime using LEFT/RIGHT/h/l to move between positions, UP/DOWN/j/k to increment/decrement values by 1, or use numerals to set the date (if valid date) in the position of the cursor.
- default time in now with zeroed hour, min, sec
- default time can be set to anything
- weekday can be turned off
- prompt can be ignored
- can specify if you want to select a date, time, or datetime
- added examples and tests

### Bug Fix
- corrected term.size() to specify the window height not with for pagination


mitsuhiko:
I played with it and it looks nice but the input for the fuzzy search is not shown anywhere and the redraw is quite flickery. Do you think this could be fixed?

slaterb1:
@mitsuhiko I added the ability to display the `search_match` string that is being filtered against. If you build the example using --release it does not flicker. I was worried about it too when I initially build it.

pksunkara:
I am willing to merge this now if you can rebase. Thanks.

slaterb1:
Updated and ready to go, thanks for taking a look

slaterb1:
@pksunkara hold up on the merge, I'm about to fix a paging bug I just discovered in all `src/select.rs`. It needs to be using `term.size().1` not `term.size().0`. terminal_size returns `{Width, Height}` not `{Height, Width}`

slaterb1:
Done, and comment is updated.

src/select.rs
```diff
@@ -24,6 +26,22 @@ pub struct Checkboxes<'a> {
     clear: bool,
     theme: &'a dyn Theme,
     paged: bool,
+    offset: usize,
+    lines_per_item: usize,
```
pksunkara, line 30: I don't agree with this `lines_per_item`. Ideally, we should calculate the str width of each item after removing ansi codes and then calculate the number of lines that item takes after text wrapping it.

src/select.rs
```diff
@@ -464,6 +509,259 @@ impl<'a> Checkboxes<'a> {
...
+        let mut size_vec = Vec::new();
+        for items in self.items.iter().as_slice() {
+            let size = &items.len();
+            size_vec.push(size.clone());
+        }
+        loop {
+            let filtered_list: Vec<&String> = self
+                .items
+                .iter()
+                .filter(|item| {
```
pksunkara, line 655: Fuzzy selection means you something different. You don't use `find` for this. For example, `cml` should be able to match `camel`.

slaterb1: We are currently using "fuzzy" selection to match across strings (i.e. `ch a` would match `match across`). I agree that the terminology is more like "AutocompleteSelection" but we are not changing the naming as we are using my fork in production.

src/select.rs
slaterb1: Clearly you didn't understand the intention here. You case would account for having variable number of items per page depending on the sizes of all of the items. I was using this for real examples of selecting from multi-line items:
Title:
Author:
Description:

slaterb1:
Thanks for the feedback, despite it taking 7 months to get. We have some disagreements in terms of logic on functions, so I'll take my fork in my own direction. At least I'll be able to update the code to modern Rust styling/conventions.

We also have differences in terms of what `absolutely broken` is. You should look into paging, as the current code doesn't actually work properly, but you guys don't capture that in any of your examples. Maybe I'll open up some BUG tickets next year.

src/select.rs
pksunkara: My case would still cover the case you wanted.

pksunkara:
Please don't hesitate to open bug tickets

pksunkara:
And the multiline example is definitely absolutely broken because I couldn't interact with it all.

slaterb1:
multiline_selection

pksunkara:
I am on mac default terminal and it wasn't working. But anyway thanks for your effort. If you don't plan to PR, I will slowly port the relevant code from your fork while making sure it is correct.

Currently, I am working on fixing a few interoperability issues on windows for console library.

slaterb1:
not going to fight it, not going to PR, but the gif was ubuntu default and I just ran it on mac default (work machine) and it still worked exactly the same.

You might want to look into the issue with your terminal, it could produce other false negatives (especially since you are furthering this library) or do yourself a favour and update to iterm2 :)

PR #33, opened 2019-09-30
Added ability to customize selection offset, use multiline selections, and use casing in FuzzySelect

slaterb1:
Now `Select`, `Checkboxes`, and `FuzzySelect` can display selections that span multiple lines and have a custom set offset.

`FuzzySelect` now includes ability to use casing in matches.


slaterb1:
meant to merge this into my fork first, sorry.

PR #34, opened 2019-10-12
2018 compatibility + fuzzy select from slaterb1

ahmed-masud:
This brings together the fuzzy selector implementation from slaterb1 and also adds 2018


deg0nz:
I just thought about implementing a fuzzy search feature for dialoguer myself. It would be nice to have this in the near future!

SirWindfield:
The move to the 2018 edition plus the new fizzy search should imo be split off into two separate PRs, making it easier to review and therefore faster to merge.

PR #35, opened 2019-11-14
Add badges

keirlawson:
I think the docs one is especially useful for discoverability.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,3 @@ readme = "README.md"
 console = ">=0.9.1, <1.0.0"
 lazy_static = "1"
 tempfile = "3"
-
-[badges]
-travis-ci = { repository = "mitsuhiko/dialoguer", branch = "master" }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -13,3 +13,6 @@ readme = "README.md"
 console = ">=0.9.1, <1.0.0"
 lazy_static = "1"
 tempfile = "3"
+
+[badges]
+travis-ci = { repository = "mitsuhiko/dialoguer", branch = "master" }
```
pksunkara, line 18: Crates.io is removing support for badges.

README.md
```diff
@@ -1,3 +1,7 @@
 # dialoguer
 
+[![Build Status](https://travis-ci.org/mitsuhiko/dialoguer.svg?branch=master)](https://travis-ci.org/mitsuhiko/dialoguer)
+[![Latest version](https://img.shields.io/crates/v/dialoguer.svg)](https://crates.io/crates/dialoguer)
+[![Documentation](https://docs.rs/dialoguer/badge.svg)](https://docs.rs/dialoguer)
```
pksunkara, line 5: Can you please use `style=flat-square` for all of them?

keirlawson: So far as I can see it seems to be only shields.io that offers that option, and they don't have a badge for docs.rs ?

PR #44, opened 2020-02-04
Resolves #43

### LINKED ISSUE #43: Match input and validator generic type
Assuming #42 gets merged, then we can always guarantee that the input value is of type `T` instead of using `&str`. So I have a request of changing the validator to receive `&T` instead of `&str`.

**Examples:**

As in, this:

```rust
Input::<i32>::new()
    .validate_with(|input: &i32| -> Result<(), &str> {
        todo!()
    });

Input::<char>::new()
    .validate_with(|input: &char| -> Result<(), &str> {
        todo!()
    });
```

Instead of:

```rust
Input::<i32>::new()
    .validate_with(|input: &str| -> Result<(), &str> {
        todo!()
    });

Input::<char>::new()
    .validate_with(|input: &str| -> Result<(), &str> {
        todo!()
    });
```

*Note `&i32` and `&char` vs `&str`  and`&str`.*

vallentin:
Ended up implementing my own change request (#43). It builds on #42, but I decided to separate the pull requests, as this change introducing minor breaking API changes, whereas #42 just fixes odd behavior.

As mentioned in the issue, this change changes input type of the validator to match the generic type of `Input`.

Personally, and in my use cases, this is a desired change, since when the parsed value is already an `i32`, then I would like to validate on that `i32` as well, instead of having to parse it a separate time.

**Example:**

```rust
Input::<i32>::new()
//      ^^^
    .validate_with(|input: &i32| -> Result<(), &str> {
//                          ^^^
        todo!()
    });

Input::<char>::new()
    .validate_with(|input: &char| -> Result<(), &str> {
        todo!()
    });
```


src/prompts.rs
```diff
@@ -222,9 +222,14 @@ where
     }
 
     /// Registers a validator.
-    pub fn validate_with<V: Validator + 'static>(&mut self, validator: V) -> &mut Input<'a, T> {
+    // pub fn validate_with<V: Validator<T> + 'static>(&mut self, validator: V) -> &mut Input<'a, T>
```
pksunkara, line 225: Unnecessary comment

vallentin:
That merge did not go as expected. However, the comment has been removed.

PR #47, opened 2020-02-24
Autocomplete selection prompt

kegesch:
This PR adds a selection prompt with autocompletion. 
When you type it automatically filters the matching items.


pksunkara:
I am closing this because fuzzy select is much better than auto complete. Also, the key bindings are not working properly.

PR #51, opened 2020-04-22
Add "Key::Unknown" handling

barrowsys:
Working on a project. Turns out i need to test that the dialogues output
a certain way. Turns out there's no way to do that because the case of
"not a terminal" is not handled! It's a quick fix, might as well. I'll
be doing the same for console.

just noticed i accidentally RustFmt'd it too. doesn't hurt but feel free to not accept those changes

### DIFF
--- a/src/select.rs
+++ b/src/select.rs
@@ -223,6 +223,13 @@ impl<'a> Select<'a> {
                     }
                     return Ok(Some(sel));
                 }
+
+                Key::Unknown => {
+                    return Err(io::Error::new(
+                        io::ErrorKind::NotConnected,
+                        "Not a terminal",
+                    ))
+                }
                 _ => {}
             }
             if sel != !0 && (sel < page * capacity || sel >= (page + 1) * capacity) {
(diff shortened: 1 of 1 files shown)


pksunkara:
I am not sure how this will help. Could you please provide how to reproduce this issue?

barrowsys:
issue being that when a dialogue is run in a non-interactive environment, it just sits there, constantly receiving "Key::Unknown" but ignoring it, providing no way to get out of the loop. with this, it will return an error instead.

pksunkara:
Yeah, I understood that part. I was asking to know how I can try dialoguer in a non-interactive environment. Can you please provide a small snippet and/or shell interactions to try?

barrowsys:
my code generates a dialogue, and i want to test that the dialogue is correct. i can redirect stdout to a string (to assert against) but without handling Key::Unknown it just sits there looping forever

pksunkara:
Okay, let me rephrase the whole conversation. Please add a test.

barrowsys:
oh ok sure thing

Issue #49, opened 2020-04-05
PasswordInput containing password buffers has to be zeroized on `drop`


fadeevab:
This is a common security problem: once password appears in the memory, it stays there unless it is explicitly erased. The problem is solved using a type which erases buffer on `drop`. 

There is a crate `rpassword` which does only one function: it prompts for a password, reads the input, but all password buffers are erased on `drop`.

@mitsuhiko I saw your blog, seems like you're alive :) Is there a sense to provide a PR so you will review it?

There are 2 options to implement it: using the `secrecy` crate which implements `SecretString` and it is pretty stable (it is used in Rust Crypto), or implement own type, but I would suggest `secrecy`. (Tech detail: it's not enough to fill the buffer with zeroes, because compiler will optimize `memset` in the end of the function as a redundant peace of code).

Thank you for a feedback!

ofadieiev:
@mitsuhiko

About this issue: only the internal **confirmation** password must be erased, so the dialoguer API stays the same. The returned password can be boxed into any zeroizable container by user. So don't worry about backward compatibility. Would you be ok about such scenario?

fadeevab:
@mitsuhiko

About this issue: only the internal confirmation password must be erased, so the dialoguer API stays the same. The returned password can be boxed into any zeroizable container by user. So don't worry about backward compatibility. Would you be ok about such scenario?

pksunkara:
@fadeevab I am willing to accept a PR.

fadeevab:
@pksunkara Cool! Give me a time to prepare a PR then.

PR #54, opened 2020-05-01
Zeroize a password confirmation buffer securely

fadeevab:
I used `secrecy` crate to automatically zeroize an internal confirmation string on drop. Another option is to use `zeroize` crate directly, but in that case need to call `pw2.zeroize()` twice in two if-else cases.

Solves #49.


fadeevab:
Tell me if additional documentation should be commited

pksunkara:
Looks okay, but I am worried about the MSRV bump to 1.39

fadeevab:
Oh, I see... I quickly checked, it doesn't seem like both crates really use rust features of 1.39+, I see they modified README (1.36->1.39) and added a few commits after that with nothing specific. Either way what is the minimum rust version of `dialoguer`?

pksunkara:
I am not sure what dialoguer's MSRV is.

fadeevab:
Nice:) Does it mean we stuck? :)

pksunkara:
Maybe for a few more rust versions?

fadeevab:
@pksunkara Ok, what I think, dialoguer's minimum rust version might be 1.31, 2018 edition, right? So let me check whether `secrecy` is compatible with this version. And as an option I will find a bit older but compatible version of secrecy. Thank you for your patience

fadeevab:
Closed in favor of PR #55

PR #55, opened 2020-05-03
Zeroize a password confirmation buffer securely

fadeevab:
`zeroize` of version "0.9.3" is used because it doesn't depend on `alloc` crate which is unstable for rustc 1.32.0 which is currently the minimum supported version for `dialoguer`.

Solves #49.


fadeevab:
I identified the minimum supported version of `rustc` for `dialoguer`: **1.32.0**

fadeevab:
Thanks, man!

PR #56, opened 2020-05-04
Zeroize password while looping

fadeevab:
One case was not covered: in a case when confirmation is wrong the loop starts again and the initial password stays in memory. Zeroize it as well, and clone the final password out.

Solves #49.

`cargo test` is OK.
`cargo run --example password` is OK.


fadeevab:
Also it will cover all error cases in between (all question marks "?"): if any error happens, the buffer will be dropped with zeroization.

PR #58, opened 2020-05-07
add new fns similar to interact that ignore unprintable characters

### LINKED ISSUE #57: term::read_key vs term::read_line
I'd like to be able to manage user input by keystroke as it is being collected, so that I can avoid displaying unprintable characters and detect ESC as a cancel action.

I'm not sure what would be the best way to accomplish this. One solution would be to add a new flavor of `read_line` to the `console` crate and a new prompt type to `dialoguer`. Is this something you think is worthwhile? It would require some coordination of PRs between `dialoguer` and `console`.

My current solution is to replace the `read_line` calls in `prompt/input`:

``` rust
// We'd like to hide non-ascii keystrokes and also offer "ESC" as a way for the user to
// cancel input... but console::term doesn't allow this. Perhaps new public functions
// could be added to the console library but this would require us to patch
// both dialoguer and console.  Instead I've moved the internals of console's
// read_line_initial_text fn here and modified it to suit our purpose.

let mut chars: Vec<char> = Vec::new();

if let Some(initial) = self.initial_text.as_ref() {
    term.write_str(initial)?;
    chars = initial.chars().collect();
}

loop {
    match term.read_key()? {
        Key::Backspace => {
            if chars.pop().is_some() {
                term.clear_chars(1)?;
            }
            term.flush()?;
        }
        Key::Escape => {
            chars = vec!['\x1B'];
            break;
        }
        Key::Enter => break,
        Key::Char(chr) => {
            if chr.is_ascii_alphanumeric() || chr.is_ascii_whitespace() {
                chars.push(chr);
                let mut bytes_char = [0; 4];
                chr.encode_utf8(&mut bytes_char);
                term.write_str(chr.encode_utf8(&mut bytes_char))?;
                term.flush()?;
            }
        }
        _ => (),
    }
}

let input = chars.iter().collect::<String>();
```

I'm sure the ESC handling sequence could be a lot cleaner but this seems to be working...

rjwalters:
Closes #57 

This replaces the input prompt's use of:

``` rust
let input = if let Some(initial_text) = self.initial_text.as_ref() {
    term.read_line_initial_text(initial_text)?
} else {
    term.read_line()?
};
```

with a new block that uses `console::term::read_key()`

This allows us to avoid printing control characters.

### DIFF
--- a/src/prompts/input.rs
+++ b/src/prompts/input.rs
@@ -145,12 +145,12 @@ where
     /// Enables the user to enter a printable ascii sequence and returns the result.
     ///
     /// The dialog is rendered on stderr.
-    pub fn get_text(&self) -> io::Result<T> {
+    pub fn interact_text(&self) -> io::Result<T> {
         self.get_text_on(&Term::stderr())
     }
 
-    /// Like `get_text` but allows a specific terminal to be set.
-    pub fn get_text_on(&self, term: &Term) -> io::Result<T> {
+    /// Like `interact_text` but allows a specific terminal to be set.
+    pub fn interact_text_on(&self, term: &Term) -> io::Result<T> {
         let mut render = TermThemeRenderer::new(term, self.theme);
 
         loop {
--- a/src/prompts/input.rs
+++ b/src/prompts/input.rs
@@ -146,7 +146,7 @@ where
     ///
     /// The dialog is rendered on stderr.
     pub fn interact_text(&self) -> io::Result<T> {
-        self.get_text_on(&Term::stderr())
+        self.interact_text_on(&Term::stderr())
     }
 
     /// Like `interact_text` but allows a specific terminal to be set.
--- a/src/prompts/input.rs
+++ b/src/prompts/input.rs
@@ -9,7 +9,7 @@ use crate::{
     validate::Validator,
 };
 
-use console::Term;
+use console::{Key, Term};
 
 /// Renders an input prompt.
 ///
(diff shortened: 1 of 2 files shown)


rjwalters:
Note that this breaks paste for input on unix terminals without also patching `console`  (see https://github.com/mitsuhiko/console/pull/62)

rjwalters:
... this actually causes problems for other types of user interaction...

``` rust
let selection = Select::with_theme(&ColorfulTheme::default())
    .default(0)
    .items(&commands)
    .interact()
    .unwrap();
```

I think it'll be better to implement as a new pub fn instead...

pksunkara:
If the other prompts are having issues, then they need to be fixed too.

src/prompts/input.rs
```diff
@@ -167,17 +167,46 @@ where
...
+            let mut chars: Vec<char> = Vec::new();
+            if let Some(initial) = self.initial_text.as_ref() {
+                term.write_str(initial)?;
+                chars = initial.chars().collect();
+            }
+            loop {
+                match term.read_key()? {
+                    Key::Backspace => {
+                        if chars.pop().is_some() {
+                            term.clear_chars(1)?;
```
pksunkara, line 182: Does this delete the initial text also?

src/prompts/input.rs
```diff
@@ -167,17 +167,46 @@ where
...
+                            chars.push(chr);
+                            let mut bytes_char = [0; 4];
+                            chr.encode_utf8(&mut bytes_char);
+                            term.write_str(chr.encode_utf8(&mut bytes_char))?;
+                            term.flush()?;
+                        }
+                    }
+                    Key::Enter => break,
+                    _ => (),
+                }
```
pksunkara, line 202: `Key::Unknown` might need to be supported.

rjwalters: ok I added this in

rjwalters:
I think trying to support a cancel action is a bad idea so I removed it - this should be its own type of prompt if that's something to be supported.

pksunkara:
Please rebase and squash the whole thing. I don't want to merge a merge commit.

rjwalters:
I think this ready now. I'm not 100% on the name.. `interact_text_only()` or `interact_without_control_chars()` are a bit longer.

src/prompts/input.rs
```diff
@@ -142,6 +142,104 @@ where
...
+                            term.clear_chars(1)?;
+                        }
+                        term.flush()?;
+                    }
+                    Key::Char(chr) => {
+                        if !chr.is_ascii_control() {
+                            chars.push(chr);
+                            let mut bytes_char = [0; 4];
+                            chr.encode_utf8(&mut bytes_char);
+                            term.write_str(chr.encode_utf8(&mut bytes_char))?;
```
pksunkara, line 192: This looks like the prompt will work with utf8 strings, right?

src/prompts/input.rs
```diff
@@ -142,6 +142,104 @@ where
...
+            }
+            loop {
+                match term.read_key()? {
+                    Key::Backspace => {
+                        if chars.pop().is_some() {
+                            term.clear_chars(1)?;
+                        }
+                        term.flush()?;
+                    }
+                    Key::Char(chr) => {
```
pksunkara, line 187: Can you add cursor left and right movements? And `Key::Delete`?

pksunkara:
I like the name `interact_text`.

pksunkara:
I actually wanted to make this input prompt similar to https://github.com/enquirer/enquirer#input-prompt with how default works. But this is a good start and I can work on it later.

src/prompts/input.rs
rjwalters: I think so, but it would be nice to have some tests.

src/prompts/input.rs
rjwalters: I think this should be possible.

rjwalters: I suppose we would need to keep track of the cursor location in `chars: Vec<char>` so it could be a little tricky.

rjwalters:
I like `interact_text()` too. I think the arrow keys, del, and fancy defaults would be great improvements now that the concept is more clear. I'd use it in its current form if it could be merged as a "minimum viable" version.

src/prompts/input.rs
pksunkara: Huh, that sounds a bit more complicated. I didn't realize it when I proposed it.

Issue #40, opened 2020-01-22
Directory Prompt for dialoguer


alicanerdogan:
Hi! Thanks for this awesome crate.

I wonder whether I can help with adding a directory prompt like this to the crate:
https://github.com/nicksrandall/inquirer-directory

I can try to add a similar feature to the crate if it is necessary and if it is not something has a very specific use.

Let me know.

pksunkara:
Yeah, please go ahead.

PR #69, opened 2020-06-11
Added callback option

BlackPhlox:
Disables the select prompt confirmation.

When a prompt is set the system does not print a confirmation after
the selection. Useful for creating cli-menu with directory display shown below. 

? Main menu > Submenu ›
❯ Back
SubSubmenu
OtherPrompt
Settings

Useful in relation to #40 (though using [jmoore914](https://github.com/jmoore914/inquirer-file-selector-prompt) and [terkelg](https://github.com/terkelg/prompts) as inspiration) which I am currently [working on](https://github.com/BlackPhlox/wte). It could be expanded to other than just the select prompt. However, it won't really make any sense to use it on the other type of prompts. Alternatively, this can be made into a separate type of prompt, which is a bit overkill IMO.


BlackPhlox:
Not sure if related but I get a [E0277] on all of the example validate_input.rs, even without my additions. I might need to create an issue on this.

BlackPhlox:
Based on terkelg's [prompts](https://github.com/terkelg/prompts) I've added callback functionality (currently only one select), necessary for the current [project](https://github.com/BlackPhlox/wte), I'm working on.

src/theme.rs
```diff
@@ -260,8 +260,8 @@ impl Default for ColorfulTheme {
             inactive_item_style: Style::new().for_stderr(),
             active_item_prefix: style("❯".to_string()).for_stderr().green(),
             inactive_item_prefix: style(" ".to_string()).for_stderr(),
-            checked_item_prefix: style("✔".to_string()).for_stderr().green(),
-            unchecked_item_prefix: style("✔".to_string()).for_stderr().black(),
+            checked_item_prefix: style("✔️".to_string()).for_stderr().green(),
+            unchecked_item_prefix: style("❌".to_string()).for_stderr().black(),
```
pksunkara, line 264: I will not be accepting the change in this diff section

src/prompts/select.rs
```diff
@@ -175,12 +179,51 @@ impl<'a> Select<'a> {
...
+    ///
+    /// When a prompt is set the system does not print a confirmation after
+    /// the selection. Useful for creating cli-menu with directory display shown below.
+    ///? Main menu > Submenu ›
+    ///❯ Back
+    /// SubSubmenu
+    /// OtherPrompt
+    /// Settings
+    ///
+    pub fn with_no_prompt_confirmation(&mut self) -> &mut Select<'a> {
```
pksunkara, line 192: I will not be accepting this new function. You can do this by building a new theme and replacing the logic inside `select_prompt_selection` and forwarding the other functions to `ColoredTheme`

src/prompts/select.rs
```diff
@@ -37,6 +37,8 @@ pub struct Select<'a> {
     default: usize,
     items: Vec<String>,
     prompt: Option<String>,
+    prompt_confirmation: bool,
+    on_render: Box<dyn FnMut(usize) -> () + 'a>,
```
pksunkara, line 41: I like this change. Probably should be renamed. Could you add this to other prompts also?

src/theme.rs
BlackPhlox: Yeah sorry, this was a rouge commit from another branch that got merged. This is resolved on the latest commit on master.

src/prompts/select.rs
BlackPhlox: Thanks, I was not aware you could do that. Maybe it would be a good idea to create documentation like a user guide for configuration and customization of prompts? Or by adding examples related to customizations etc. Just a thought.

src/prompts/select.rs
BlackPhlox: Sure thing, I was about to do that anyway as I need some of the other prompts for my project. As mentioned, the name is based on terkelg's [prompts](https://github.com/terkelg/prompts). Can't come up with a better name atm.

BlackPhlox:
Sorry for the git-mess. The password prompt version of on_render is put on hold for now due to &mut self. Documentation and cleanup of examples in progress.

pksunkara:
I am thinking about this. Why exactly did you need `on_render`? Wouldn't customizing the theme solve your issue?

pksunkara:
> On_render is generally useful

Can you please explain a use case? I can't think of one. Would appreciate it.

BlackPhlox:
> > On_render is generally useful
> 
> Can you please explain a use case? I can't think of one. Would appreciate it.

Anything that requires real-time validation (validates as you write) or updates as you go through a selection. In my use-case writing to a JSON file as you are navigating a prompt. For more use cases just look at some of the projects using [terkelg's prompts](https://github.com/terkelg/prompts/network/dependents), the number of repo's should speak for itself.

BlackPhlox:
I think this PR is too long and convoluted, once I've made some progress on my part (documentation and test) I'll create a new PR that cuts everything to the bone.

pksunkara:
There is validation already available. And to do things based on the selection you already get the result from the prompt which you can use.

I still don't see an use case that isn't solved by existing API. Maybe when you create the new PR you could try to show a small example of the issue you are having.

BlackPhlox:
I'll see what I can do, but I don't think you understand. I must stress that "real-time validation" is a key difference. Currently, validation only occurs when you have confirmed your prompt. By real-time, I mean during navigation and prior to confirmation of the prompt.

Here is a visual example shown below. When I go through the selection of images, each time a new update (and therefore a new selection) occurs, a different method is run setting the current selection to the new image of the background. I hope this helps. 

wt-editor-cli-showcase

PR #70, opened 2020-06-18
Add an option to wait for the user to hit Enter before proceeding

### LINKED ISSUE #66: Confirmation should wait for Enter key to be pressed
Currently the confirmation proceeds immediately on a single key being input, which isn't really inline with conventional CLI design. I'd expect it to wait for Enter to be pressed even when giving an explicit answer.

abreis:
This PR adds an optional behavior to `Confirm` in which the user must type out their option and hit the Enter key before proceeding. This is settable with a `.wait_for_newline(bool)` method on its builder.

The two major differences are:
1. The user can now type strings, so we accept "yes" and "no" along with "y" and "n", as well as uppercase and mixed-case variations.
2. When the input is not a valid choice, the prompt is re-rendered to the user (because we're on a new line).

Example:
```
$ ./test
Do you want to continue? [Y/n] asd
Do you want to continue? [Y/n] ye
Do you want to continue? [Y/n] y
Looks like you want to continue
```

The old behavior is still the default so as not to break existing code that may be interacting with it programatically. In a future semver-breaking release, I'm of the opinion that this new behavior should be made the default.

Closes #66.

### DIFF
--- a/src/prompts/confirm.rs
+++ b/src/prompts/confirm.rs
@@ -23,6 +23,7 @@ pub struct Confirm<'a> {
     prompt: String,
     default: bool,
     show_default: bool,
+    wait_for_newline: bool,
     theme: &'a dyn Theme,
 }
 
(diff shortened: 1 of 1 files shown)


pksunkara:
I have decided not to make this behaviour default since it's really simple to enable.

PR #73, opened 2020-07-24
Add option to Confirm that forces user to enter 'y' or 'n' to continue

### LINKED ISSUE #64: Confirm with no default
In a CLI-based wizard, it's quite easy to accidentally hit the Enter key an accidental extra time and mistakenly select a default confirmation you wanted to change.

As a feature, it would be nice not only to change the default, but to disable a default all-together, such that you *must* type a "y" or "n" before the prompt returns.

edwardwli:
Fixes #64

### DIFF
--- a/examples/confirm.rs
+++ b/examples/confirm.rs
@@ -3,11 +3,10 @@ extern crate dialoguer;
 use dialoguer::Confirm;
 
 fn main() {
-    println!("with confirm");
+    println!("disable default");
     if Confirm::new()
         .with_prompt("continue?")
-        .confirm(true)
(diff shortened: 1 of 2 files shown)


pksunkara:
I was rethinking this recently. I think the fix should be a new API called `disable_default()` which disables the `default`.

Or change the `default` to `Option<bool>` and keep it as `None`.

edwardwli:
I changed `confirm()` with `disable_default()`.

pksunkara:
I have removed the need for this code by making confirm prompt not have any default value by default. This use case will be satisfied with that.

PR #79, opened 2020-08-26
feat: add fuzzy single select prompt

mainrs:
Superseeds #34.

There is sadly one problem that I can't resolve. I wanted to add the search term to the prompt so that users can see what they typed. But I can't find a way to delete the old prompt. My attempt is commented out inside the loop.

When I include the commented out code, I still get a new prompt each time I type a character.
Follow-up with a working GIF.


SirWindfield:
In the example I type `ice`. Since `Ice` matches exactly, it is shown first. 
fuzzy

SirWindfield:
Applying this patch does display the search prompt. As far as I understood, there is no performance impact. Preserve prompt does also delete all lines but keeps the prompt intact. Manually calling `clear` does delete the prompt as well.

```patch
diff --git a/src/prompts/fuzzy.rs b/src/prompts/fuzzy.rs
index 1e6859c..024460a 100644
--- a/src/prompts/fuzzy.rs
+++ b/src/prompts/fuzzy.rs
@@ -158,15 +158,11 @@ impl<'a> FuzzySelect<'a> {
         let matcher = fuzzy_matcher::skim::SkimMatcherV2::default();
 
         loop {
-            // TODO: resolve
-            // if !search_term.is_empty() {
-            //     render.clear_preserve_prompt(&size_vec)?;
-            //     match &self.prompt {
-            //         Some(prompt) => render.select_prompt(&format!("{}: {}", prompt, search_term))?,
-            //         None => render.select_prompt(&format!("> {}", search_term))?,
-            //     };
-            // }
-            // ENDTODO
+            render.clear()?;
+            match &self.prompt {
+                Some(prompt) => render.input_prompt_selection(prompt, &search_term)?,
+                None => render.select_prompt(&format!("> {}", search_term))?,
+            };
 
             // Maps all items to a tuple of item and its match score.
             let mut filtered_list = self.items.iter()
@@ -258,7 +254,6 @@ impl<'a> FuzzySelect<'a> {
             if filtered_list.len() > 0 && (sel < page * capacity || sel >= (page + 1) * capacity) {
                 page = sel / capacity;
             }
-            render.clear_preserve_prompt(&size_vec)?;
         }
     }
 }
\ No newline at end of file

```

SirWindfield:
I'll open another issue for this, but working on the code base the first time I had a hard time understanding some logic. Some parts have a rather high cognitive complexity and without comments they are hard to grasp. This should be opened as a new issue though.

Maybe some internal documentation would be nice as well. For example the render methods could include a short example on what the output looks like exactly. Ideally, `write_formatted_str` would actually return the string and not delegate the call to `self.term`. That way, the examples could be even run using `rustdoc` tests.

SirWindfield:
Fixed the prompt in the latest commit. Here are two screenshots, one with and one without a prompt text. The commit message does explain my changes to `Theme` to accomplish this:
image
sc

Cargo.toml
```diff
@@ -19,3 +19,4 @@ console = "0.11.2"
 lazy_static = "1"
 tempfile = "3"
 zeroize = "0.9.3"
+fuzzy-matcher = "0.3.5"
```
pksunkara, line 22: Let's put this behind an optional feature flag.

src/lib.rs
```diff
@@ -20,7 +20,7 @@
 pub use console;
 pub use edit::Editor;
 pub use prompts::{
-    confirm::Confirm, input::Input, multi_select::MultiSelect, password::Password, select::Select,
+    confirm::Confirm, input::Input, fuzzy::FuzzySelect, multi_select::MultiSelect, password::Password, select::Select,
```
pksunkara, line 23:
```suggestion
    confirm::Confirm, input::Input, fuzzy_select::FuzzySelect, multi_select::MultiSelect, password::Password, select::Select,
```

src/prompts/mod.rs
```diff
@@ -1,4 +1,5 @@
 pub mod confirm;
+pub mod fuzzy;
```
pksunkara, line 2:
```suggestion
pub mod fuzzy_select;
```

src/theme.rs
```diff
@@ -522,6 +536,32 @@ impl Theme for ColorfulTheme {
 
         write!(f, "{} {}", details.0, details.1)
     }
+
+    /// Formats an input prompt after selection.
+    fn format_fuzzy_prompt(
```
pksunkara, line 541: Please use the name `fuzzy_select` everywhere instead of just `fuzzy`.

pksunkara (changes requested):

deg0nz:
Hi, just a quick question about the return value of the `interact()` functions in this PR:
Is there a specific reason, why this here is returning a `String`? 
The other select-type prompts in dialoguer return `usize` as the position of the selected item(s) in the input array (`select`) or an array of the positions (`multi_select`) respectively. I was just starting to include the current state of this PR in my code and was expecting a `usize`, which confused me a little bit 😅 .

kegesch:
@deg0nz I was confused as well. Returning the index would coherent.
@SirWindfield If there are more items than free lines in the terminal, the line with the prompt and the search string is not displayed anymore and thus hard to see what was typed.

pksunkara:
The second issue is a general issue with dialoguer and need to be fixed in a more abstract way.

src/prompts/fuzzy_select.rs
```diff
@@ -0,0 +1,252 @@
...
+    /// The index of the selected item. None if the user
+    /// cancelled with Esc or 'q'.
+    /// The dialog is rendered on stderr.
+    pub fn interact_opt(&self) -> io::Result<Option<String>> {
+        self._interact_on(&Term::stderr(), true)
+    }
+
+    /// Like `interact` but allows a specific terminal to be set.
+    pub fn interact_on(&self, term: &Term) -> io::Result<String> {
+        self._interact_on(term, false)?.ok_or(io::Error::new(
```
pksunkara, line 122: `ok_or_else(|| io::Error`

src/prompts/fuzzy_select.rs
```diff
@@ -0,0 +1,252 @@
...
+    fn _interact_on(&self, term: &Term, allow_quit: bool) -> io::Result<Option<String>> {
+        let mut page = 0;
+        let mut capacity = self.items.len();
+        let mut search_term = String::new();
+
+        if self.paged {
+            capacity = (term.size().0 as usize) / self.lines_per_item - self.offset;
+        }
+
+        let pages = (self.items.len() / capacity) + 1;
```
pksunkara, line 143: `let pages = (self.items.len() as f64 / capacity as f64).ceil() as usize;`

src/prompts/fuzzy_select.rs
```diff
@@ -0,0 +1,252 @@
...
+        ))
+    }
+
+    /// Like `interact` but allows a specific terminal to be set.
+    pub fn interact_on_opt(&self, term: &Term) -> io::Result<Option<String>> {
+        self._interact_on(term, true)
+    }
+
+    /// Like `interact` but allows a specific terminal to be set.
+    fn _interact_on(&self, term: &Term, allow_quit: bool) -> io::Result<Option<String>> {
```
pksunkara, line 134: Out should be `io::Result<Option<usize>>`

src/prompts/fuzzy_select.rs
```diff
@@ -0,0 +1,252 @@
...
+        let matcher = fuzzy_matcher::skim::SkimMatcherV2::default();
+
+        loop {
+            render.clear()?;
+            render.fuzzy_select_prompt(self.prompt.as_deref(), &search_term)?;
+
+            // Maps all items to a tuple of item and its match score.
+            let mut filtered_list = self.items.iter()
+                .map(|item| (item, matcher.fuzzy_match(item, &search_term)))
+                .filter_map(|(item, score)| match score {
```
pksunkara, line 163: `|(item, score)| score.map(|s| (item, s))`

src/prompts/fuzzy_select.rs
```diff
@@ -0,0 +1,252 @@
...
+
+            // Renders all matching items, from best match to worst.
+            filtered_list.sort_unstable_by(|(_, s1), (_, s2)| s2.cmp(&s1));
+            capacity = filtered_list.len();
+            if self.paged {
+                capacity = (term.size().0 as usize) / self.lines_per_item - self.offset;
+            }
+
+            for (idx, (item, _)) in filtered_list.iter().enumerate().skip(page * capacity).take(capacity) {
+                render.select_prompt_item(item, idx == sel)?;
```
pksunkara, line 177: Please create a new `fuzzy_select_prompt_item` and corresponding functions in the theme

src/prompts/fuzzy_select.rs
```diff
@@ -0,0 +1,252 @@
...
+                    if sel == !0 {
+                        sel = 0;
+                    } else {
+                        sel = (sel as u64 + 1).rem(filtered_list.len() as u64) as usize;
+                    }
+                }
+                Key::Escape => {
+                    if allow_quit {
+                        if self.clear {
+                            term.clear_last_lines(filtered_list.len())?;
```
pksunkara, line 191: `term.flush()?` needed here.

src/prompts/fuzzy_select.rs
```diff
@@ -0,0 +1,252 @@
...
+                }
+                Key::ArrowUp if filtered_list.len() > 0 => {
+                    if sel == !0 {
+                        sel = filtered_list.len() - 1;
+                    } else {
+                        sel = ((sel as i64 - 1 + filtered_list.len() as i64)
+                            % (filtered_list.len() as i64)) as usize;
+                    }
+                }
+                Key::ArrowLeft => {
```
pksunkara, line 204: In `Input`, arrow left & right are used to move the cursor in the input which we need here to for the search term. We need to solve this issue.

src/prompts/fuzzy_select.rs
```diff
@@ -0,0 +1,252 @@
...
+                        sel = page * capacity;
+                    }
+                }
+
+                Key::Enter if filtered_list.len() > 0 => {
+                    if self.clear {
+                        render.clear()?;
+                    }
+                    if let Some(ref prompt) = self.prompt {
+                        render.input_prompt_selection(prompt, &filtered_list[sel].0)?;
```
pksunkara, line 230: Please create new `fuzzy_select_prompt_selection` and corresponding functions in theme.

src/prompts/fuzzy_select.rs
```diff
@@ -0,0 +1,252 @@
...
+            filtered_list.sort_unstable_by(|(_, s1), (_, s2)| s2.cmp(&s1));
+            capacity = filtered_list.len();
+            if self.paged {
+                capacity = (term.size().0 as usize) / self.lines_per_item - self.offset;
+            }
+
+            for (idx, (item, _)) in filtered_list.iter().enumerate().skip(page * capacity).take(capacity) {
+                render.select_prompt_item(item, idx == sel)?;
+            }
+
```
pksunkara, line 179: Need to move cursor to the prompt and flush here.

src/prompts/fuzzy_select.rs
```diff
@@ -0,0 +1,252 @@
...
+/// Renders a selection menu that user can fuzzy match to reduce set.
+pub struct FuzzySelect<'a> {
+    default: usize,
+    items: Vec<String>,
+    prompt: Option<String>,
+    clear: bool,
+    theme: &'a dyn Theme,
+    paged: bool,
+    offset: usize,
+    lines_per_item: usize,
```
pksunkara, line 17: I don't think we need this. Ideally dialoguer should figure out the number of lines and intelligently do display them.

src/prompts/fuzzy_select.rs
```diff
@@ -0,0 +1,252 @@
+use crate::theme::{SimpleTheme, Theme, TermThemeRenderer};
+use crate::Select;
+use std::io;
+use console::{Term, Key};
+use fuzzy_matcher::FuzzyMatcher;
+use std::ops::Rem;
```
pksunkara, line 6: Please merge the imports.

pksunkara (changes requested):
Sorry, I wanted to fix the `Confirm` prompt first before looking at this which is why it took me quite a bit of time to review.

ttys3:
what 's the status of this issue ? could it be merged to the master?

ttys3:
`examples/fuzzyselect.rs` has dos file format, it should be unix.
`src/prompts/fuzzy_select.rs` has dos file format, it should be unix.

SirWindfield:
I will not continue to work on this PR. If someone wants to take over, feel free to do so!

deg0nz:
I would like to take over. I just don't know how to contribute to this particular PR. I've never taken over a PR before.

@SirWindfield can you assign it to me somehow? Or should I do something?

pksunkara:
You can make a new PR.

ttys3:
I think I could try to make a new PR. but currently my fork is just get this PR merged, I need to look into @pksunkara 's review message carefully and make some change to the code, and then make a new PR.

deg0nz:
Hey, @pksunkara 

I started to apply your suggestions in my fork and came pretty far. Could you please answer the questions I add to your comments above and explain some more stuff?

src/prompts/fuzzy_select.rs
```diff
@@ -0,0 +1,252 @@
...
+        self
+    }
+
+    /// Sets number of lines paged offset includes
+    pub fn offset(&mut self, val: usize) -> &mut FuzzySelect<'a> {
+        self.offset = val;
+        self
+    }
+
+    /// Enables or disables paging
```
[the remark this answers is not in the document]
deg0nz: Where can I find an example for "real" paging?

src/prompts/fuzzy_select.rs
deg0nz: Should `term.clear_last_lines(filtered_list.len())?;` be replaced here, or should `term.flush()?` be added to the if clause? 

(I currently replaced it and it seems to work fine)

src/prompts/fuzzy_select.rs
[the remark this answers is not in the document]
pksunkara: To test, make your terminal small in height and have more options than the number of lines. fzf would be a good example but not perfect.

src/prompts/fuzzy_select.rs
pksunkara: `flush` doesn't clear. We actually need to clear the last lines.

src/prompts/fuzzy_select.rs
deg0nz: I have a question about this regarding the paging function:

Since we use the arrows left & right for navigating the cursor now, what should we do about the paging function? It is not navigable anymore (horizontally). 
Should we just assume, that the fuzzy search narrows the results down so we only need to navigate with the up & down keys?

deg0nz: I have an additional problem: 

Due to the fact, that we use an input combined with a select prompt, the terminal cursor is rendered at the bottom of the select prompt and not in the input prompt. (Of course, because there is the last character that was written to the terminal)

Do you have any idea, how I can render the cursor to the input prompt?

Here is a GIF of what I mean:
fuzzy_cursor

The only thing that comes to my mind is implementing a styleable "cursor" by myself that highlights the current position in the input string and add it to the rendering function.

deg0nz: Just in case you didn't get a notification, here's a ping @pksunkara :)

pksunkara: I think there are some terminal functions that allow you to move the cursor up and down lines. This has been a fundamental issue in the library for all prompts that I have been meaning to fix.

pksunkara: > Since we use the arrows left & right for navigating the cursor now, what should we do about the paging function? It is not navigable anymore (horizontally).

Not sure what to do. We need to fix the paging problem anyway when we have multiple pages of options for normal select prompt. But until then you can assume just one page since fuzzy narrows it down.

PR #81, opened 2020-09-08
Remove 'static restriction on Input<T> validate_with() and fix tests

jamesmcm:
See https://github.com/jamesmcm/vopono/issues/19 for how this caused issues previously.

Also fixed current tests to run for current codebase.

Note #43 means that the validators now need to take `&String` not `&str` (since the type signature dictates `&T`). This is a bit painful.

This is because for the Input<T>, T is the owned type i.e. String, so then the closure (on `&T`) must accept &String and cannot accept &str. I tried to work around this with AsRef but it won't work unless we change the T to be the base type i.e. &str and then use ToOwned to make it return the owned type (we could even use Cow). But this is a lot of work and would change all instantiations of `Input<T>`.

### DIFF
--- a/src/prompts/multi_select.rs
+++ b/src/prompts/multi_select.rs
@@ -5,15 +5,18 @@ use crate::theme::{SimpleTheme, TermThemeRenderer, Theme};
 use console::{Key, Term};
 
 /// Renders a multi select prompt.
-/// 
+///
 /// ## Example usage
 /// ```rust,no_run
+/// # fn test() -> Result<(), Box<dyn std::error::Error>> {
 /// use dialoguer::MultiSelect;
-/// 
-/// let items = vec![("Option 1", true), ("Option 2", false)];
+///
+/// let items = vec!["Option 1", "Option 2"];
 /// let chosen : Vec<usize> = MultiSelect::new()
 ///     .items(&items)
 ///     .interact()?;
+/// # Ok(())
+/// # }
 /// ```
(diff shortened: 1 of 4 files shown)


src/prompts/multi_select.rs
```diff
@@ -5,15 +5,18 @@ use crate::theme::{SimpleTheme, TermThemeRenderer, Theme};
...
-/// 
+///
 /// ## Example usage
 /// ```rust,no_run
+/// # fn test() -> Result<(), Box<dyn std::error::Error>> {
 /// use dialoguer::MultiSelect;
-/// 
-/// let items = vec![("Option 1", true), ("Option 2", false)];
+///
+/// let items = vec!["Option 1", "Option 2"];
```
pksunkara, line 14: This would be wrong.

jamesmcm: What was it meant to do?

That test was failing to build because `(&str, bool)` doesn't implement `ToString` - required here: https://github.com/mitsuhiko/dialoguer/blob/076a96c169aab99f11781185d7be3c73bc2e7976/src/prompts/multi_select.rs#L94

pksunkara: Oh, nvm. I got it wrong. I was thinking about https://github.com/mitsuhiko/dialoguer/pull/81/files#diff-dfda20335a245e26af9f9bd3c7ba7eb9R103

pksunkara (approved):

PR #88, opened 2020-10-16
fix args in vim -N -u NONE

lengyijun:
It's an experiment pr to fix https://github.com/mitsuhiko/dialoguer/issues/87


pksunkara (approved):

lengyijun:
Thanks! 
But actually, I only test it on Linux and I don't have windows and macs so I can't test them. I feel that this issue requires a bit more testing anyway.

PR #90, opened 2020-10-19
Upgrade zeroize dependency

jplatte:


pksunkara:
> `zeroize` of version "0.9.3" is used because it doesn't
depend on `alloc` crate which is unstable for rustc 1.32.0
which is currently the minimum supported version for `dialoguer`.

jplatte:
Where is that quote from? Should probably be a comment in `Cargo.toml`.

pksunkara:
It was a quote from when the dep was added.

PR #91, opened 2020-10-19
Remove implicit dependency on zeroize_derive

jplatte:
Followup to #90.


jplatte:
If this could be released in a new patch release that would be great :)

PR #96, opened 2020-11-15
Allow FnMut for validate

SOF3:
This allows mutable closures for validation, allowing stateful
validators.

This breaks BC. Input::interact*() now take &mut self instead of &self.
However, this shall not break most usres,
since (I think) users usually use Input in chained syntax
or call one of the &mut methods (like with_prompt) separately,
which would require a `let mut` anyway.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "dialoguer"
 description = "A command line prompting library."
-version = "0.8.0"
+version = "0.7.1"
 edition = "2018"
 authors = [
 	"Armin Ronacher <armin.ronacher@active-4.com>",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "dialoguer"
 description = "A command line prompting library."
-version = "0.7.1"
+version = "0.8.0"
 edition = "2018"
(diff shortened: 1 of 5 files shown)


pksunkara:
Can you please add an example?

SOF3:
What example would you like? This just widens the type accepted by `validate_with`. It doesn't add any features.

pksunkara:
We do not have an example with validate in `examples` folder. Since you added the abitlity for stateful validators, I was asking you to please create an example with a stateful validator.

SOF3:
I appended this to the existing validator example.

pksunkara:
Everything looks good. Can you please add an entry to changelog and I can merge this.

SOF3:
Updated. Also bumped minor version since it's a BC break.

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "dialoguer"
 description = "A command line prompting library."
-version = "0.7.1"
+version = "0.8.0"
```
pksunkara, line 4: Can you revert this line? I will change it when releasing.

pksunkara (changes requested):

pksunkara (approved):

PR #101, opened 2021-01-22
Allow confirm with no choice

### LINKED ISSUE #100: interact_opt and interact_on_opt for Confirm
Sometimes, a user might be presented with a `(y/n)` prompt, and be undecided (i.e. neither "yes" nor "no"). It would be helpful if this library added `interact_opt` and `interact_on_opt` to the `Confirm` struct, which would return `Option<bool>`.

spenserblack:
Adds `Confirm::interact_opt` and `Confirm::interact_on_opt` to handle situations where the user
wishes to answer neither "yes" nor "no".

Resolves #100 

***

# Bug(s) to fix

- [x] When `default(true)`, `wait_for_newline(true)`, and `interact_opt` called on a `Confirm`, then `Esc` does not clear the final value -- `Esc` followed by `Enter` will return `Some(true)`, not `None`.

### DIFF
--- a/examples/confirm.rs
+++ b/examples/confirm.rs
@@ -58,6 +58,7 @@ fn main() {
 
     match Confirm::with_theme(&ColorfulTheme::default())
         .with_prompt("Do you really really really really really want to continue?")
+        .default(true)
         .wait_for_newline(true)
         .interact_opt()
         .unwrap()
--- a/examples/confirm.rs
+++ b/examples/confirm.rs
@@ -45,4 +45,14 @@ fn main() {
     } else {
         println!("nevermind then :(");
     }
+
+    match Confirm::with_theme(&ColorfulTheme::default())
+        .with_prompt("Do you really really really really want to continue?")
+        .interact_opt()
+        .unwrap()
+    {
+        Some(true) => println!("Looks like you want to continue"),
+        Some(false) => println!("nevermind then :("),
+        None => println!("Ok, we can start over later"),
+    }
 }
--- a/examples/confirm.rs
+++ b/examples/confirm.rs
@@ -55,4 +55,15 @@ fn main() {
         Some(false) => println!("nevermind then :("),
         None => println!("Ok, we can start over later"),
     }
+
+    match Confirm::with_theme(&ColorfulTheme::default())
+        .with_prompt("Do you really really really really really want to continue?")
+        .wait_for_newline(true)
+        .interact_opt()
+        .unwrap()
+    {
+        Some(true) => println!("Looks like you want to continue"),
+        Some(false) => println!("nevermind then :("),
+        None => println!("Ok, we can start over later"),
+    }
 }
--- a/src/prompts/confirm.rs
+++ b/src/prompts/confirm.rs
@@ -205,7 +205,9 @@ impl<'a> Confirm<'a> {
                         value = Some(false);
                     }
                     Key::Enter => {
-                        value = value.or(self.default);
+                        if !allow_quit {
+                            value = value.or(self.default);
+                        }
 
                         if value.is_some() || allow_quit {
                             rv = value;
--- a/src/prompts/confirm.rs
+++ b/src/prompts/confirm.rs
@@ -137,6 +137,10 @@ impl<'a> Confirm<'a> {
     /// # }
     /// ```
     pub fn interact_on(&self, term: &Term) -> io::Result<bool> {
+        self._interact_on(term, false)
+    }
+
+    fn _interact_on(&self, term: &Term, allow_quit: bool) -> io::Result<bool> {
         let mut render = TermThemeRenderer::new(term, self.theme);
 
         let default_if_show = if self.show_default {
--- a/src/prompts/confirm.rs
+++ b/src/prompts/confirm.rs
@@ -182,6 +182,9 @@ impl<'a> Confirm<'a> {
                             continue;
                         }
                     }
+                    Key::Escape | Key::Char('q') if allow_quit => {
+                        value = None;
+                    }
                     _ => {
                         continue;
                     }
@@ -199,6 +202,7 @@ impl<'a> Confirm<'a> {
                     Key::Char('y') | Key::Char('Y') => Some(true),
                     Key::Char('n') | Key::Char('N') => Some(false),
                     Key::Enter if self.default.is_some() => Some(self.default.unwrap()),
+                    Key::Escape | Key::Char('q') if allow_quit => None,
                     _ => {
                         continue;
                     }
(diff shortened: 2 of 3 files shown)


src/theme.rs
```diff
@@ -40,12 +40,23 @@ pub trait Theme {
...
-            write!(f, "{} {}", &prompt, if selection { "yes" } else { "no" })
+        match selection {
+            Some(selection) => {
+                if prompt.is_empty() {
+                    write!(f, "{}", if selection { "yes" } else { "no" })
+                } else {
+                    write!(f, "{} {}", &prompt, if selection { "yes" } else { "no" })
+                }
+            }
+            None => {
```
pksunkara, line 53: This needs more thought.

src/prompts/confirm.rs
```diff
@@ -158,25 +195,28 @@ impl<'a> Confirm<'a> {
...
 
                         if let Some(val) = value {
-                            rv = val;
+                            rv = Some(val);
                             break;
                         } else {
                             continue;
                         }
                     }
+                    Key::Escape | Key::Char('q') => {
```
pksunkara, line 217: You need `if allow_quit` here to check.

src/prompts/confirm.rs
```diff
@@ -158,25 +195,28 @@ impl<'a> Confirm<'a> {
...
+                    Key::Char('n') | Key::Char('N') => {
                         value = Some(false);
                     }
-                    '\n' | '\r' => {
+                    Key::Enter => {
                         value = value.or(self.default);
 
                         if let Some(val) = value {
-                            rv = val;
+                            rv = Some(val);
```
pksunkara, line 211: I am not sure about this line. Did you check that the previous behavior remains the same?

pksunkara (changes requested):

src/prompts/confirm.rs
spenserblack: I'll push up a commit soon that changes this behavior slightly. After some thought, I think that, if waiting for newline, <kbd>Esc</kbd> or <kbd>q</kbd> should clear the selection, but continue waiting for newline.

src/theme.rs
spenserblack: Is the problem the coding style or the behavior?

The behavior makes sense to me -- if there is no prompt and no answer, there is nothing to write to screen. Otherwise, write only the prompt. Should a placeholder like `"?"` or `"no answer"` be written to the screen?

src/prompts/confirm.rs
spenserblack: Thank you for finding that! It was foolish of me to forget that :facepalm: :sweat_smile:

src/prompts/confirm.rs
spenserblack: I pushed up a commit that now writes `n/a` if the answer was neither `yes` nor `no`. After thinking about it a bit, it's much better :smiley:

Also, I was able to make the change much more readable than what I previously did :laughing:

src/theme.rs
spenserblack: I pushed up a commit that now writes `n/a` if the answer was neither `yes` nor `no`. After thinking about it a bit, I believe it's much better :smiley:

Also, I think my changes in this file are now much more readable than what I previously did :laughing:

spenserblack:
@pksunkara Thank you for your feedback! I pushed up some commits that address your comments.

examples/confirm.rs
```diff
@@ -45,4 +45,25 @@ fn main() {
...
+        Some(true) => println!("Looks like you want to continue"),
+        Some(false) => println!("nevermind then :("),
+        None => println!("Ok, we can start over later"),
+    }
+
+    match Confirm::with_theme(&ColorfulTheme::default())
+        .with_prompt("Do you really really really really really want to continue?")
+        .wait_for_newline(true)
+        .interact_opt()
+        .unwrap()
```
pksunkara, line 63: In this example, when pressing `q` after prompt show (without pressing `Enter` yet) should show `n/a`. It shows `yes` or `no` when `y` or `n` is pressed.

pksunkara (changes requested):
I am still not sold on the `n/a` being shown. Is there any prior work in other languages/programs we can look at? How about `cancel`?

examples/confirm.rs
spenserblack: Just to confirm what it should look like. Is this what you're envisioning for `interact_opt` with `wait_for_newline(true)`?

###### Prompt displayed, no user input
```
? Do you really really really really really want to continue? (y/n) ›
```


###### User presses `Esc` or `q`
```
? Do you really really really really really want to continue? (y/n) › n/a
```

###### User presses `Enter` after `Esc` or `q`
```
? Do you really really really really really want to continue? (y/n) › cancel
```

pksunkara: yeah, but `cancel` instead of `n/a` in second scenario too.

spenserblack: Got it. That makes sense to me for the first and third scenarios (I chose `n/a` to limit to 3 characters, but I have no strong opinion on what it should be), but I have a question about the second scenario.

***

What would this mean for the `default` method?

Currently, there are 3 states for a confirm prompt:
- nothing (no input ***or*** `q` or `Esc` input)
- `no` (`n` input)
- `yes` (`y` input)

and `default` lets you set the initial state to either `Yes` or `No`.

If `cancel` is displayed when `Esc` or `q` are pressed, then there are 4 states:
- nothing (no input)
- `no` (`n` input)
- `yes` (`y` input)
- `cancel` (`q` or `Esc` input)

`default` is used to set the initial state. Currently it takes a `bool`, so the initial state can be either `yes` or `no`. If `cancel` is a state that can be displayed before pressing `Enter`, then I don't think `default` taking a `bool` would be sufficient. In my opinion, it would then need to take either an `Option<bool>` or an enum with `No`, `Yes`, and `Cancel` variants. This would be a breaking change.

If `q` and `Esc` reverts to the "nothing" state, then I think a user not calling `default` before `interact_opt` can be assumed to mean that they want *no value* to be the default value. But if `cancel` is displayed in the second scenario, and `default` continues to take a `bool`, then what would the user not calling `default` mean? Would it mean they want *no default value*, or that the default value is `cancel`?

spenserblack:
# Note to self

Found that, when default is set in an optional confirm that waits for newline, cancel is ignored -- if initial state is `yes`, `Esc` is pressed, then `Enter`, the final value will be `Some(true)`, not `None`.

examples/confirm.rs
pksunkara: What about removing `cancel` and `n/a` and just not displaying anything for `None` state?

spenserblack: >just not displaying anything for `None` state?

Isn't that the current behavior as the PR currently is?

pksunkara: After giving it some thought, let's remove `cancel` from the end too. Basically no `cancel` or `n/a` ever.

spenserblack: Done

PR #104, opened 2021-02-19
Use `console::measure_text_width` to generate size_vec in select

CircuitCoder:
This allows using ANSI escape codes in select items.

Previously line width are measured using `str::len`, which causes text with ANSI escape codes to be measured longer than they really are. Subsequently when clearing printed lines, additional lines can be cleared.

### DIFF
--- a/src/prompts/multi_select.rs
+++ b/src/prompts/multi_select.rs
@@ -159,8 +159,8 @@ impl<'a> MultiSelect<'a> {
             .flat_map(|i| i.split('\n'))
             .collect::<Vec<_>>()
         {
-            let size = &items.len();
-            size_vec.push(*size);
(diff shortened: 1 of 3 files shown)


pksunkara:
I would prefer if we integrate https://github.com/mgeisler/textwrap. That library is very high quality (I can attest to) and is being used in clap, and we personally don't need to maintain this text_width related logic.

CircuitCoder:
OK I'll look into that when I have some time.

Right now, the measured width is only used for clearing printed lines in select/sort/mult_select. If we switch to `textwrap`, are we going to change the way we render the options too?

pksunkara:
Yeah, we need to handle multi-lines text anyway normally.

PR #109, opened 2021-03-13
Update dependency: console: 0.13.0 -> 0.14.0

matthiasbeyer:


matthiasbeyer:
CI failure because installing lib32gcc-5-dev failed to be installed, not because of this code.

PR #117, opened 2021-04-22
Add must_use lint to interaction functions

matthiasbeyer:
This is a draft because there's a problem with that patch:

* If one uses the `Result` part, the return value seems to be used and the compiler does not emit a warning for the unused `bool` value.

Maybe we can fix this somehow, I've [asked the community](https://users.rust-lang.org/t/must-use-lint-for-functions-returning-result/58746) how such a thing would be possible.

### DIFF
--- a/src/prompts/confirm.rs
+++ b/src/prompts/confirm.rs
@@ -115,6 +115,7 @@ impl<'a> Confirm<'a> {
     /// Otherwise function discards input waiting for valid one.
     ///
     /// The dialog is rendered on stderr.
+    #[must_use = "Value must be used to have an effect"]
     pub fn interact(&self) -> io::Result<bool> {
         self.interact_on(&Term::stderr())
     }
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
Would you accept a patch to change the interface as suggested [by steffahn in the discussion from the users.rust-lang.org forum](https://users.rust-lang.org/t/must-use-lint-for-functions-returning-result/58746/2)?

src/prompts/confirm.rs
```diff
@@ -115,6 +115,7 @@ impl<'a> Confirm<'a> {
     /// Otherwise function discards input waiting for valid one.
     ///
     /// The dialog is rendered on stderr.
+    #[must_use = "Value must be used to have an effect"]
```
pksunkara, line 118: Let's rename all to just "value must be used".

pksunkara (commented):
What is your proposal for the interface change?

matthiasbeyer:
@pksunkara to not return a `bool` but rather an enum that can be used to determine whether the answer to the question was "yes" or "no", with a `must_use`, just like [steffahn proposed](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9981a3a0fdfa0833749e3c3d584684d3).

That could even implement `Into<bool>`, with a `must_use` annotation on the `fn into(self) -> bool` implementation.

pksunkara:
No, I don't think the interface should be changed to accomodate a bug in the compiler which might get fixed anytime.

matthiasbeyer:
What bug?

pksunkara:
The bug that compiler is not warning that bool result is not used.

matthiasbeyer:
Can you link an issue? Because I don't think that this is a bug in the compiler, because the Result is actually used!

pksunkara:
Sorry, I misunderstood a little bit.

I think the point still stands that I don't want to create an enum and complicate the API just to make sure people who are using this library didn't forget something.

PR #119, opened 2021-05-10
fix example build when editor feature isn't enabled

mcginty:
I didn't realize it when I submitted the pull request, sorry.

### DIFF
--- a/examples/editor.rs
+++ b/examples/editor.rs
@@ -1,10 +1,15 @@
-use dialoguer::Editor;
-
+#[cfg(feature = "editor")]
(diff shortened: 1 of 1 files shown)


pksunkara (changes requested):
The better thing to do is to use https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-required-features-field.

mcginty:
Thanks for the tip! Never heard of that cargo option before.

pksunkara:
Also needed for password and fuzzy select features.

PR #120, opened 2021-05-10
make zeroize dependency optional

mcginty:
Similar to the `editor` feature change, making it simple for people to not have to bring in dependencies for features they don't use.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,11 +16,11 @@ documentation = "https://docs.rs/dialoguer"
 readme = "README.md"
 
 [features]
-default = ["editor"]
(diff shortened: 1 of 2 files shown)


pksunkara (changes requested):
While I do like the change on a personal level, I reckon we should not make this a feature it all in the first place because we want to encourage best practices regarding security for handling password prompts. IMHO having it as a default feature is not enough of an encouragement.

pksunkara:
That would be great.

pksunkara:
But honestly, how much of a difference is zeroize giving in size and compile times? I remember it being a really small crate.

PR #122, opened 2021-06-04
Add example targets

JunichiSugiura:
to easily interact with.

e.g.
```
cargo run --example <example-name>
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,3 +24,39 @@ console = "0.14.1"
 lazy_static = "1"
 tempfile = { version = "3", optional = true }
 zeroize = "1.1.1"
+
+[[example]]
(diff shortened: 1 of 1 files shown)


pksunkara:
They currently work without needing to add them as targets because of default cargo auto discovery.


```
→ cargo run --example input
   Compiling regex-syntax v0.6.23
   Compiling terminal_size v0.1.16
   Compiling regex v1.4.5
   Compiling console v0.14.1
   Compiling dialoguer v0.8.0 (/Users/pksunkara/Coding/mitsuhiko/dialoguer)
    Finished dev [unoptimized + debuginfo] target(s) in 9.06s
     Running `target/debug/examples/input`
? Your name › ⏎          
```

PR #126, opened 2021-08-11
Add support for optional history processing to `Input`

CraZySacX:
* Added support for optional history processing to `Input`.  

Up/Down arrow keys are read in `interact_text_on` and allow for history to be read via a user supplied function.
History is written after successful input validation via the same user supplied function.

### DIFF
--- a/src/history.rs
+++ b/src/history.rs
@@ -1,0 +1,18 @@
+use std::{fmt::Display, str::FromStr};
+
+/// Trait for history handling
+pub trait History<T> {
+    /// This is called with the current position that should
+    /// be read from history.  The `pos` represents the number
+    /// of times the `Up`/`Down` arrow key has been pressed.
+    /// This would normally be used as an index to some sort
+    /// of vector.  If the `pos` does not have an entry, [`None`](Option::None)
+    /// should be returned.
+    fn read(&self, pos: usize) -> Option<String>;
+    /// This is called with the next value you should store
+    /// in history at the first location.  Normally history
+    /// is implemented as a FIFO queue.
+    fn write(&mut self, val: &T)
+    where
+        T: Clone + FromStr + Display;
+}
--- a/src/history.rs
+++ b/src/history.rs
@@ -1,4 +1,4 @@
-use std::{fmt::Display, str::FromStr};
+use std::fmt::Display;
 
 /// Trait for history handling
 pub trait History<T> {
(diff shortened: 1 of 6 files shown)


pksunkara (changes requested):
Can you please add an example called `history.rs`?

jasonozias:
Will do.  And seems to have cause some build failures.  I'll look into those as well.

pksunkara:
While I would appreciate you fixing the failures, I think they are unrelated.

src/prompts/input.rs
```diff
@@ -45,6 +45,7 @@ pub struct Input<'a, T> {
     theme: &'a dyn Theme,
     permit_empty: bool,
     validator: Option<Box<dyn FnMut(&T) -> Option<String> + 'a>>,
+    history: Option<Box<dyn FnMut(usize, Option<&T>) -> Option<String> + 'a>>,
```
pksunkara, line 48: While I am okay with this, what do you think about a `History` struct that has `read` and `write` fns? And all of this should be gated behind `history` feature.

jasonozias: I initially had something like that setup, but consolidated down to one.   I can go that route again no problem.   And I'll add the gate as well.

jasonozias:
Went with a `History` trait and added the feature flags.

src/history.rs
```diff
@@ -0,0 +1,18 @@
...
+    /// This would normally be used as an index to some sort
+    /// of vector.  If the `pos` does not have an entry, [`None`](Option::None)
+    /// should be returned.
+    fn read(&self, pos: usize) -> Option<String>;
+    /// This is called with the next value you should store
+    /// in history at the first location.  Normally history
+    /// is implemented as a FIFO queue.
+    fn write(&mut self, val: &T)
+    where
+        T: Clone + FromStr + Display;
```
pksunkara, line 17: Do you need `FromStr`?

src/prompts/input.rs
```diff
@@ -292,6 +359,103 @@ where
...
 
+    #[cfg(feature = "history")]
+    fn write_history(&mut self, value: &T) {
+        if let Some(history) = &mut self.history {
+            history.write(value);
+        }
+    }
+
+    #[cfg(not(feature = "history"))]
+    fn write_history(&mut self, _value: &T) {}
```
pksunkara, line 370: I don't think you need this functions. You can put the `cfg` on statements in the `_interact` itself.

pksunkara (changes requested):

src/history.rs
jasonozias: Removed.

src/prompts/input.rs
jasonozias: Moved back to being gated inline.

pksunkara (approved):

PR #129, opened 2021-08-21
Add max page sized to select

justinbarclay:
Often times when I am building a select at the CLI, I don't want it to take up the entire terminal and instead I'd like it to take up a set size.

I've hacked around in dialoguer a little bit to get what I want working and I am wondering if the maintainers would like this feature backported to them?


pksunkara:
Blocked on #130 to be done first

pksunkara (changes requested):
Blocked

pksunkara:
@justinbarclay Since the paging refactor has been done, do you want to update this? cc @deg0nz

justinbarclay:
Hey, thanks for the ping. I'll try to take a look at this on the weekend 🙂

Issue #125, opened 2021-08-11
Fix paging for all prompts


sassman:
origin of this issue is https://github.com/cargo-generate/cargo-generate/issues/400 where a test example is provided and little more context is given.

deg0nz:
Hi,

I hope, @pksunkara meant "increase usability for paging" by "Fix paging for all prompts". :)

(I looked into the code of `cargo-generate` and the error that happens in cargo-generate/cargo-generate#400, is that they use a loooooooong list without paging, which makes the select prompt unusable)

Anyways, did a quick&dirty demo with the select prompt to suggest how we could increase the usability with paging. 
Maybe that is what you had in mind...

Here's what I added:

- Prompt is always shown on every page
- A page indicator is printed next to the prompt

Here's how it looks with the select example:
![better_paging](https://user-images.githubusercontent.com/6370295/129628296-4388063d-b69e-4e0c-9c7c-5ace1397e021.gif)

I also tried it with `cargo-generate` with the repo described in cargo-generate/cargo-generate#400. It looks like this:
![better_paging_cargo_generate](https://user-images.githubusercontent.com/6370295/129631199-cb4daa70-f5d7-452e-bba4-4abcae046f46.gif)

If this is what you had in mind, I could start implementing this for all prompts that need paging and make a PR to discuss the details.

**Another thought here:**
IMHO dialoguer should figure out for itself if it needs to activate paging or not. It could just check, if `selections.len() > capacity` and activate paging automatically.

deg0nz:
D'uh. I somehow managed to delete my comment (I can't see it anymore) 🤦‍♀️. So here we go again:

I hope, @pksunkara meant, "Increase paging usability" with "Fix paging for all prompts". :)

I looked into the code of `cargo-generate` regarding cargo-generate/cargo-generate#400. What happens there is, that they use a looooong list _without_ paging, which makes the select prompt unusable.

Anyways, I did a quick&dirty demo to suggest how we could increase the usability with paging. Here's what it could look like with the select prompt:

![better_paging](https://user-images.githubusercontent.com/6370295/129633526-1c967847-b01e-4e2d-bb94-528fd0ff50af.gif)


I also did it with the example mentioned in cargo-generate/cargo-generate#400, which then would look like this:

![better_paging_cargo_generate](https://user-images.githubusercontent.com/6370295/129633587-26b199f8-8ac6-43b8-8eed-4b9a72b8c7ac.gif)

[...]

PR #130, opened 2021-08-29
Fix paging

deg0nz:
This addresses #125 

The following things have been addressed:

* I extracted all the functions regarding the paging functionality into an own module to reduce code redundancy
* The module detects if paging is needed and (de)activates paging accordingly
* The module is able to update terminal information via an `update()` function. This allows the rendering to adjust to terminal resizing. (But since the `match term.read_key()?` seems to block the rendering loop, the healing will only be achieved after the next key press)
* All prompts that are related to item lists (except `fuzzy-select` for now) are adjusted to use the new paging module.
    * The prompt rendering functions have an additional parameter to be able to print the paging information

Currently the whole thing behaves like this when the terminal is resized:

fix_paging

### DIFF
--- a/examples/select.rs
+++ b/examples/select.rs
@@ -6,27 +6,52 @@ fn main() {
         "Vanilla Cupcake",
         "Chocolate Muffin",
         "A Pile of sweet, sweet mustard",
+        "Foo",
+        "Bar",
+        "Baz",
+        "Mustard",
+        "Cream",
+        "Banana",
+        "Chocolate",
+        "Flakes",
+        "Corn",
+        "Cake",
+        "Tarte",
+        "Clear glaze",
+        "Vanilla",
+        "Hazelnut",
+        "Flour",
+        "Sugar",
+        "Salt",
+        "Potato",
+        "French Fries",
+        "Pizza",
+        "Mousse au chocolat",
+        "Brown sugar",
+        "Blueberry",
+        "Blanc",
     ];
 
     let selection = Select::with_theme(&ColorfulTheme::default())
         .with_prompt("Pick your flavor")
         .default(0)
+        .paged(true)
         .items(&selections[..])
         .interact()
         .unwrap();
 
     println!("Enjoy your {}!", selections[selection]);
 
-    let selection = Select::with_theme(&ColorfulTheme::default())
-        .with_prompt("Optionally pick your flavor")
-        .default(0)
-        .items(&selections[..])
-        .interact_opt()
-        .unwrap();
+    // let selection = Select::with_theme(&ColorfulTheme::default())
+    //     .with_prompt("Optionally pick your flavor")
+    //     .default(0)
+    //     .items(&selections[..])
+    //     .interact_opt()
+    //     .unwrap();
 
-    if let Some(selection) = selection {
-        println!("Enjoy your {}!", selections[selection]);
-    } else {
-        println!("You didn't select anything!");
-    }
+    // if let Some(selection) = selection {
+    //     println!("Enjoy your {}!", selections[selection]);
+    // } else {
+    //     println!("You didn't select anything!");
+    // }
 }
--- a/examples/select.rs
+++ b/examples/select.rs
@@ -6,52 +6,27 @@ fn main() {
         "Vanilla Cupcake",
         "Chocolate Muffin",
         "A Pile of sweet, sweet mustard",
-        "Foo",
-        "Bar",
-        "Baz",
-        "Mustard",
-        "Cream",
-        "Banana",
-        "Chocolate",
-        "Flakes",
-        "Corn",
-        "Cake",
-        "Tarte",
-        "Clear glaze",
-        "Vanilla",
-        "Hazelnut",
-        "Flour",
-        "Sugar",
-        "Salt",
-        "Potato",
-        "French Fries",
-        "Pizza",
-        "Mousse au chocolat",
-        "Brown sugar",
-        "Blueberry",
-        "Blanc",
     ];
 
     let selection = Select::with_theme(&ColorfulTheme::default())
         .with_prompt("Pick your flavor")
         .default(0)
-        .paged(true)
         .items(&selections[..])
         .interact()
         .unwrap();
 
     println!("Enjoy your {}!", selections[selection]);
 
-    // let selection = Select::with_theme(&ColorfulTheme::default())
-    //     .with_prompt("Optionally pick your flavor")
-    //     .default(0)
-    //     .items(&selections[..])
-    //     .interact_opt()
-    //     .unwrap();
+    let selection = Select::with_theme(&ColorfulTheme::default())
+        .with_prompt("Optionally pick your flavor")
+        .default(0)
+        .items(&selections[..])
+        .interact_opt()
+        .unwrap();
 
-    // if let Some(selection) = selection {
-    //     println!("Enjoy your {}!", selections[selection]);
-    // } else {
-    //     println!("You didn't select anything!");
-    // }
+    if let Some(selection) = selection {
+        println!("Enjoy your {}!", selections[selection]);
+    } else {
+        println!("You didn't select anything!");
+    }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -27,6 +27,7 @@ pub use prompts::{
     sort::Sort,
 };
 pub use validate::Validator;
+pub use paging::Paging;
 
 #[cfg(feature = "fuzzy-select")]
 pub use prompts::fuzzy_select::FuzzySelect;
@@ -38,3 +39,4 @@ mod history;
 mod prompts;
 pub mod theme;
 mod validate;
+mod paging;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,7 +26,7 @@ pub use prompts::{
     confirm::Confirm, input::Input, multi_select::MultiSelect, select::Select, sort::Sort,
 };
 pub use validate::Validator;
-pub use paging::Paging;
+use paging::Paging;
 
 #[cfg(feature = "fuzzy-select")]
 pub use prompts::fuzzy_select::FuzzySelect;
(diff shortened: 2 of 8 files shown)


examples/select.rs
```diff
@@ -6,27 +6,52 @@ fn main() {
         "Vanilla Cupcake",
         "Chocolate Muffin",
         "A Pile of sweet, sweet mustard",
+        "Foo",
+        "Bar",
+        "Baz",
+        "Mustard",
+        "Cream",
+        "Banana",
```
pksunkara, line 14: Make a different example called `paging.rs`

src/lib.rs
```diff
@@ -27,6 +27,7 @@ pub use prompts::{
     sort::Sort,
 };
 pub use validate::Validator;
+pub use paging::Paging;
```
pksunkara, line 30: Why do we need to export it as part of lib?

src/prompts/multi_select.rs
```diff
@@ -47,13 +49,15 @@ impl<'a> MultiSelect<'a> {
...
     }
 
     /// Enables or disables paging
-    pub fn paged(&mut self, val: bool) -> &mut MultiSelect<'a> {
-        self.paged = val;
+    #[deprecated(
+        since = "0.9.0",
+        note = "Activating paging has no effect anymore. Paging will be activated automatically if needed."
+    )]
+    pub fn paged(&mut self, _val: bool) -> &mut MultiSelect<'a> {
```
pksunkara, line 60: Let's delete the function directly, we dont need deprecation warnings.

src/prompts/select.rs
```diff
@@ -79,15 +79,17 @@ impl<'a> Select<'a> {
...
     /// Enables or disables paging
     ///
     /// Paging is disabled by default
-    pub fn paged(&mut self, val: bool) -> &mut Select<'a> {
-        self.paged = val;
+    #[deprecated(
+        since = "0.9.0",
+        note = "Activating paging has no effect anymore. Paging will be activated automatically if needed."
+    )]
+    pub fn paged(&mut self, _val: bool) -> &mut Select<'a> {
```
pksunkara, line 92: Let's delete this.

src/prompts/select.rs
```diff
@@ -244,30 +246,17 @@ impl<'a> Select<'a> {
...
-        let capacity = if self.paged {
-            term.size().0 as usize - 1
-        } else {
-            self.items.len()
-        };
-
-        let pages = (self.items.len() as f64 / capacity as f64).ceil() as usize;
-
         let mut render = TermThemeRenderer::new(term, self.theme);
+        let mut paging = Paging::new(term, self.items.len());
```
pksunkara, line 257: Move this line above `render`.

src/prompts/sort.rs
```diff
@@ -48,13 +47,15 @@ impl<'a> Sort<'a> {
...
     }
 
     /// Enables or disables paging
-    pub fn paged(&mut self, val: bool) -> &mut Sort<'a> {
-        self.paged = val;
+    #[deprecated(
+        since = "0.9.0",
+        note = "Activating paging has no effect anymore. Paging will be activated automatically if needed."
+    )]
+    pub fn paged(&mut self, _val: bool) -> &mut Sort<'a> {
```
pksunkara, line 58: Let's delete this.

src/prompts/sort.rs
```diff
@@ -108,21 +108,10 @@ impl<'a> Sort<'a> {
...
-        let capacity = if self.paged {
-            term.size().0 as usize - 1
-        } else {
-            self.items.len()
-        };
-
-        let pages = (self.items.len() as f64 / capacity as f64).ceil() as usize;
-
         let mut render = TermThemeRenderer::new(term, self.theme);
+        let mut paging = Paging::new(term, self.items.len());
```
pksunkara, line 112: Move this above `render`

src/theme.rs
```diff
@@ -105,6 +105,18 @@ pub trait Theme {
         self.format_prompt(f, prompt)
     }
 
+    /// Formats a select prompt with pages.
+    fn format_select_prompt_paged(
```
pksunkara, line 109: Unusued function. Let's delete this.

src/paging.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    /// Returns number of pages
+    pub fn pages(&self) -> usize {
+        self.pages
+    }
+
+    /// Returns current page
+    pub fn current_page(&self) -> usize {
+        self.current_page
+    }
```
pksunkara, line 84: The above 4 functions can be just direct member access by making members `pub`. Let's delete them.

src/paging.rs
```diff
@@ -0,0 +1,128 @@
...
+
+impl<'a> Paging<'a> {
+    pub fn new(term: &'a Term, items_len: usize) -> Paging<'a> {
+        // Subtract -2 because we need space to render the prompt, if paging is active
+        let capacity = term.size().0 as usize - 2;
+        let pages = (items_len as f64 / capacity as f64).ceil() as usize;
+
+        Paging {
+            term,
+            current_term_size: term.size(),
```
pksunkara, line 28: Let's cache `term.size()` into a var in this function instead of calling it twice.

src/paging.rs
```diff
@@ -0,0 +1,128 @@
...
+            active: pages > 1,
+            // Set transition initially to true to trigger prompt rendering for inactive paging on start
+            activity_transition: true,
+        }
+    }
+
+    /// Updates all internal based on the current terminal size and cursor position
+    pub fn update(&mut self, cursor_pos: usize) -> io::Result<()> {
+        if self.current_term_size != self.term.size() {
+            self.current_term_size = self.term.size();
```
pksunkara, line 42: Let's cache `term.size()` into a var in this function instead of calling it twice.

src/paging.rs
```diff
@@ -0,0 +1,128 @@
...
+        &mut self,
+        mut render_prompt: F,
+    ) -> io::Result<()> {
+        let mut paging_info = None;
+
+        if self.active {
+            paging_info = Some((self.current_page + 1, self.pages));
+            render_prompt(paging_info)?;
+        } else if self.activity_transition {
+            render_prompt(paging_info)?;
```
pksunkara, line 99:
```suggestion
            render_prompt(None)?;
```

We don't need to declare `paging_info` at the beginning of this function.

src/paging.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    /// Returns current page
+    pub fn current_page(&self) -> usize {
+        self.current_page
+    }
+
+    /// Renders a prompt when the following conditions are met:
+    /// * Paging is active
+    /// * Transition of the paging activity happened (active -> inactive / inactive -> active)
+    pub fn render_prompt<F: FnMut(Option<(usize, usize)>) -> io::Result<()>>(
```
pksunkara, line 89: Let's put `F: ...` in `where` instead of here.

pksunkara (changes requested):
Good progress. I want to see if we can improve the following:

1. Show pages after the prompt instead of before? or maybe something else?
2. Re render when term size changes instead of being blocked.

examples/select.rs
deg0nz: Done

I'll comment all your comment with "Done" to make sure, I didn't forget anything

src/lib.rs
deg0nz: We don't need this. Just a typo.

Done

src/prompts/multi_select.rs
deg0nz: done

src/prompts/select.rs
deg0nz: Done

src/prompts/select.rs
deg0nz: Done

src/prompts/sort.rs
deg0nz: Done

src/prompts/sort.rs
deg0nz: Done

src/theme.rs
deg0nz: Done

src/paging.rs
deg0nz: Done

src/paging.rs
deg0nz: Done

src/paging.rs
deg0nz: Done

src/paging.rs
deg0nz: Done

src/paging.rs
deg0nz: Done

deg0nz:
I applied all your review suggestions.

I will look into dynamic resizing and position of paging information and ping you, when I have progress.

deg0nz:
Ok, I changed the prompt so that the paging info right to the prompt as in my inital example in #125.

**I gave the "automatic re-rendering on terminal resize" thing some thought:**

Since the `interact_on()`'s `loop`is blocked by the `match term.read_key()`, there is no real chance to check for the terminal size change without having an extra busy loop that observes the values of `term.size()`. 
I thought about adding an observer to `console::Term`, but `console` itself just uses the `terminal_size` crate which is polling the information from the OS. Since there is no real setter for the terminal size inside `Term`, we cannot add an observer for setting the value.

So here's an idea:

We implement an (optional) setting, where a new thread is spawned that permanently loops over `term.size()` and raises an event, if the value has changed. This event is then received by `interact_on` and a re-rendering of the output is triggered.

Possible caveats:

I seems that `terminal_size` gets the terminal size from `STDOUT`'s file descriptor on unix. I don't know how much load a busy `loop` on such a file descriptor would create. We could just try it out and see what happens...

We could also ask the `terminal_size` maintainers if they have an idea, how one could implement an observer over these values.

/ping @pksunkara

pksunkara:
Looks good. Linting job is failing.

> We could also ask the terminal_size maintainers if they have an idea, how one could implement an observer over these values.

I would appreciate if you can follow up on that and we can look to improve this feature in the future. But for now, I will merge this as soon as linting is fixed.

deg0nz:
Linting is fixed.

> I would appreciate if you can follow up on that and we can look to improve this feature in the future.

Will do.

PR #132, opened 2021-09-01
Refactor: treat `password` module as an optional feature

Some-Dood:
This PR continues the work done on #120. However, instead of disabling `zeroize` (which implies insecure inputs), this approach disables the `password` module altogether. As discussed in #120, we must enforce that the user must have either a **secure** password input or no password input at all—"all or nothing", so to speak.

For backwards-compatibility, the `password` feature is enabled by default.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,14 +16,15 @@ documentation = "https://docs.rs/dialoguer"
 readme = "README.md"
 
 [features]
-default = ["editor", "history"]
+default = ["editor", "history", "password"]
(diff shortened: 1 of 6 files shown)


pksunkara (commented):
Looks good but please fix the lint errors.

Some-Dood:
No problemo! Everything should be fixed now. 👍

Though, please disregard the large diff; most of it were from `cargo fmt`.

pksunkara (approved):

PR #133, opened 2021-09-01
Refactor: loosen trait bounds and separate various `impl` blocks into bound-specific chunks

Some-Dood:
# Loosened Trait Bounds
Previously, the `read` method of the `History` trait required that `T` implements `Clone` and `Display`. However, nothing in the `History` trait (by itself) actually requires this. That is, the trait bounds are `impl`-specific, which makes the `History` trait less compatible with other types. This PR addresses the "over-restriction" by removing the trait bounds altogether. The `Clone + Display` bound is delegated to the specific `impl` blocks that require them.

Similarly, the `Err` associated type for the `Validate` trait has been loosened. Nothing in the `Validate` trait (by itself) actually requires that `Err` implements `Debug` and `Display`. Thus, the trait bounds have been delegated to the `impl` blocks that require them.

# Bound-specific Chunks
While I was refactoring the trait bounds, I also separated various `impl` blocks into _bound-specific chunks_. That is, for example in the `Input` struct, all methods that require `Debug` have been moved to their own `impl<T: Debug> Input<'_, T>` block. Meanwhile, all methods that require `Debug + ToString` have their own `impl<T: Debug + ToString> Input<'_, T>` block as well. See the diff for more examples of this pattern.

# Using the `Self` Alias
Finally, as part of the cleanup, various structs have also been refactored to make use of their respective `Self` aliases. This makes the code more concise by removing the explicit lifetime annotations.


Some-Dood:
Sure! I'll let you know when everything's ready.

Some-Dood:
@pksunkara Alrighty, everything seems to be in order. Please do let me know if there are any issues. Thanks! 🚀

src/prompts/sort.rs
```diff
@@ -133,7 +125,7 @@ impl<'a> Sort<'a> {
     ///
     /// ## Examples
     /// ```rust,no_run
-    /// use dialoguer::Select;
+    /// use dialoguer::MultiSelect;
```
pksunkara, line 128:
```suggestion
    /// use dialoguer::Sort;
```

pksunkara: Edit the usage in the line below too.

src/prompts/input.rs
```diff
@@ -51,49 +47,26 @@ pub struct Input<'a, T> {
...
 }
 
-impl<'a, T> Default for Input<'a, T>
-where
-    T: Clone + FromStr + Display,
-    T::Err: Display + Debug,
-{
-    fn default() -> Input<'a, T> {
-        Input::new()
+impl<T> Default for Input<'_, T> {
```
pksunkara, line 50:
```suggestion
impl<T> Default for Input<'static, T> {
```

src/prompts/input.rs
```diff
@@ -197,29 +146,79 @@ where
...
+        }));
+
+        self
+    }
+}
 
+impl<T> Input<'_, T>
+where
+    T: Clone + ToString + FromStr,
+    <T as FromStr>::Err: Debug + ToString,
```
pksunkara, line 220: Not sure why this one has `Debug` while the below `impl` block doesn't have it.

pksunkara (changes requested):

PR #136, opened 2021-10-05
Define unified interface for selecting/sorting prompts and changes as follows ↓

Animeshz:
Overview of the changes:

* Properly define unified interface (`interact` | `interact_opt` | `interact_on` | `interact_on_opt`) on all the select/sorting variants.
* Return `Ok(None)` when multi-select is quit using `Esc` or `q`, instead of default values provided, so the consumer of library know when user quit and when used `Enter`.
* Properly provide `#[inline]` to those interact functions, to depend on `_interact_on`, previously were only on 1 of 4 in select.rs's funciton and not present in any function in fuzzy_select.rs for some reason.
* Unify the documentation.
* Add examples which were missing from `interact_on` and `interact_on_opt` on `multi-select` and `sort`.

### DIFF
--- a/src/prompts/select.rs
+++ b/src/prompts/select.rs
@@ -168,18 +168,21 @@ impl<'a> Select<'a> {
 
     /// Enables user interaction and returns the result.
     ///
-    /// Similar to [interact_on](#method.interact_on) except for the fact that it does not allow selection of the terminal.
+    /// The user can select the items with the space bar or enter and the index of selected item will be returned.
     /// The dialog is rendered on stderr.
-    /// Result contains index of a selected item.
+    /// Result contains `Vec<index>` if user selected one of items using 'Enter'.
+    /// This unlike [interact_opt](#method.interact_opt) does not allow to quit with 'Esc' or 'q'.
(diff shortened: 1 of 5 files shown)


src/prompts/select.rs
```diff
@@ -168,18 +168,21 @@ impl<'a> Select<'a> {
 
     /// Enables user interaction and returns the result.
     ///
-    /// Similar to [interact_on](#method.interact_on) except for the fact that it does not allow selection of the terminal.
+    /// The user can select the items with the space bar or enter and the index of selected item will be returned.
     /// The dialog is rendered on stderr.
-    /// Result contains index of a selected item.
+    /// Result contains `Vec<index>` if user selected one of items using 'Enter'.
```
pksunkara, line 173: `Vec<index>`?

pksunkara (changes requested):

pksunkara:
On all the doc strings for these methods, can you please make sure keypressed are surrounded with `'`? I have seen some `enter` which should be `'Enter'`.

Issue #137, opened 2021-10-07
Feature Request: Interceptor for MultiSelect


Animeshz:
It can help in validating a few related items, on-time, in the multi-select case.

pksunkara:
You would need to expand on this. I am not sure what exactly you mean.

Animeshz:
It could be used to deny the selection & unselection of few items based on few conditions. Like for example listing packages linearly for silent installation, if one depends on another, we don't want to uncheck the one that is dependency of another and ticked 🤔, or we also want to check the dependency while selecting the packages, etc.

I'm not too sure if its viable or not, cuz currently I could do it by adding a new confirm prompt that it was not selected and is a dependency, do you want to add, etc. If its really complex and not very useful, maybe close it :))

PR #140, opened 2021-10-14
Add map method for changing inputted value

yoshrc:
The motivation for this feature is almost same as doc comment of `map` method.

Although my need is sufficed by this PR, the implementation still has some design issues.

### Probably `mutate` method is necessary

I think it is very common usage for `map` method to appending some suffix to the inputted `String`. Writing a mutator instead of a mapper makes this task easy a bit because return value can be omitted.

### Execution order of mappers and validators

Imagine the following code

```
let value = Input::new()
    .validate(v1)
    .map(m)
    .validate(v2)
    .interact_text()
    .unwrap();
```

The user of `dialoguer` surely expect that `v1` validates `input` (the raw inputted value) and `v2` validates `m(input)`. However, in my implementation, both `v1` and `v2` validate `m(input)`.

To achieve this, mappers, mutators and validators should be treated as a unified "transformer" type, whose type is `FnMut(T) -> Result<T, Error>`.

### Mapping to a different type

There is no reason to restrict return type of mappers to the same type of input.

To achieve this, map's signature should be

```
pub fn map<U>(self, mut mapper: impl FnMut(T) -> U + 'a) -> Input<'a, U>
```

I think it is very misleading that only `map` method takes `self` and other ones takes `&mut self`, so it is good choice to change `Input` into [a consuming builder](https://doc.rust-lang.org/1.0.0/style/ownership/builders.html#consuming-builders:). However this is a huge breaking change for `dialoguer`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,3 +28,6 @@ lazy_static = "1"
 tempfile = { version = "3", optional = true }
 zeroize = { version = "1.1.1", optional = true }
 fuzzy-matcher = { version = "0.3.7", optional = true }
+
+[dev-dependencies]
+regex = "1.5.4"
--- a/examples/input.rs
+++ b/examples/input.rs
@@ -1,4 +1,5 @@
 use dialoguer::{theme::ColorfulTheme, Input};
+use regex::Regex;
 
 fn main() {
     let input: String = Input::with_theme(&ColorfulTheme::default())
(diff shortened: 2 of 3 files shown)


yoshrc:
> Does this update the text on terminal with the new mapped value?

Yes. [This change](https://github.com/mitsuhiko/dialoguer/pull/140/files#diff-e791088022b7342278b5923ecbf02de36460add2840b1894114b8faa73b22723R437) does.

> How does this interact with #137?

I think #137 should be achieved by some validation mechanism, not by mapping. Specifically, a user passes a validation function whose type is `Fn(&[bool]) -> Result<(), String>`, and it checks selections whenever the user changes which one is selected. If it returns `Err`, the user cannot determine the selections and shows the message in `Err` such as "When you select A, select B too."

So I don't think my feature need specific consideration for #137.

Issue #149, opened 2021-11-10
screen reader support in select and multiselect.


alex19EP:
to improve accessibility of select and multi select widgets a terminal cursor should be set at entries which currently
highlighted with "❯".

this will greatly improve the usability of applications which depends on this library for blind people.

thank you.

pksunkara:
@alex19EP Can you please go into more details? And have you audited the other prompts too? I am afraid I don't know anything about this.

alex19EP:
> @alex19EP Can you please go into more details?

ok. screen reader are a program which reads with a speech synthesizer or outputs wire braille display information from the screen, this allows blind people to use the computer.

what is important in our case, the screen reader can read the terminal by lines and symbols, as well as pronounce the information that appears, however, when the program has a menu or dialog with several buttons, then the screen reader needs to somehow find out which menu item or button is currently selected. the cursor is used for this, the screen reader will read the line or character pointed to by the cursor.

this crait uses the ❯ symbol to indicate the selected item. if, together with this symbol, the terminal cursor is placed on the selected object, then the screen reader will be able to voice which object is currently selected.

> And have you audited the other prompts too? I am afraid I don't know anything about this.

are there any other prompts where multiple selections are available?

pksunkara:
Yeah, there is a sort prompt. I am not sure how normal select prompt works too in this situation. Which is why I would appreciate if you would audit all the prompts thoroughly.

alex19EP:
ok will look at it tomorrow.

alex19EP:
normal select works as I described above, but I cannot understand how multi select works. I see a ✔ symbol in front of each element. this is normal it can be read by a screen reader, but I do not see an indication of the selected item, maybe it is marked with a color?
also I can’t understand how the status is shown about whether the item is selected. the ✔ symbol does not change when pressing the space bar and enter confirms selection.

for sort I see the same problems. it is not clear how the selected element is indicated, as well as sorting.

pksunkara:
> but I do not see an indication of the selected item, maybe it is marked with a color?

Yeah, it is marked with color. Known issue #114

alex19EP:
OK. first, we need to show the cursor in all prompts where there is a choice of several elements, and then we can figure out how to present the data in a form accessible to screen readers.

p.s. can I somehow change the theme in the examples? I want to see how a not colored theme is rendered.

pksunkara:
Try the simple theme instead of the colored theme. You should be able to see it in docs.rs

PR #152, opened 2021-11-23
use distinct symbols for selection

### LINKED ISSUE #114: ColorfulTheme's MultiSelect basically can't be used with an empty environment
For reasons that are yet unclear to me, ColorfulTheme becomes monochrome when being used with an empty environment (`env -i [binary]`).

This, in turn, makes its `MultiSelect`, that exclusively relies on colors to identify which items are selected, unusable.

Would it make sense to have ColorfulTheme automatically detect when it's being used in a way that will make colors not work, and have it fall back to SimpleTheme? If not, how would you suggest one to handle this scenario?

Anyway, thank you for indicatif and dialoguer, they're both awesome crates :)

alex19EP:
this fixes #114 and also improves accessability related to #149


pksunkara:
Superseded by #275

PR #153, opened 2021-12-05
tidy clippy linting

danieleades:
- add clippy config file aligned with MSRV set in CI
- move `deny warnings` from the CI script to a compiler flag to ensure consistent behaviour during development and CI
- address a few pedantic lints


pksunkara (commented):
We want the clippy to be run on stable, not msrv.

danieleades:
> We want the clippy to be run on stable, not msrv.

that's not what the clippy config does. setting the MSRV prevents clippy from suggesting changes that use syntax or features which are not compatible with the MSRV.

pksunkara (approved):

pksunkara:
Thanks! ❤️

PR #154, opened 2021-12-05
use 2018-style module layout

danieleades:


pksunkara:
The original is intentional. We want all related code to be contained in prompts directory instead of dividing it into two separate places. This is how a lot of people use rust too.

PR #156, opened 2021-12-05
add dependabot config

danieleades:


pksunkara:
Dependabot adds more work than the problems it solves. I always upgrade dependencies before releases appropriately depending on their changes.

PR #157, opened 2021-12-05
put example behind conditional flag

danieleades:
the 'completion' example returns import errors if the 'completion' feature is not used. This PR places the whole example behind this feature flag to prevent the import error


danieleades:
i do wonder if this is more of a `rust-analyzer` bug, than a `dialoguer` problem. The required features for the example *are* listed in `Cargo.toml`...

pksunkara:
Yup, it is a bug with the IDE. Try running `cargo run --example completion` and you will see a proper error message.

PR #158, opened 2021-12-05
add coverage workflow

danieleades:
adds a coverage workflow using `codecov`.

Coverage in this repo is pretty low, so the intention would be for this workflow to be purely informational, and provide an opportunity to push PRs to at least *improve* coverage over time.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,29 @@
+on:
+  push:
+    branches: [master]
+  pull_request:
+
+name: Code Coverage
+
+jobs:
+  coverage:
+    name: coverage
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout source
+        uses: actions/checkout@v2
+
+      - name: Install nightly toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          profile: minimal
+          override: true
+
+      - name: Run cargo-tarpaulin
+        uses: actions-rs/tarpaulin@v0.1
+        with:
+          args: --all-features --tests --doc
+
+      - name: Upload to codecov.io
+        uses: codecov/codecov-action@v2.1.0


pksunkara:
Unfortunately until we have test suite (integration which actually tests the terminal), we can't write any tests and thus no coverage.

1. https://github.com/taiki-e/cargo-llvm-cov is better than tarpualin (due to experience)
2. We will probably not go with codecov.io and instead I can ask @mitsuhiko to setup coveralls.

danieleades:
> Unfortunately until we have test suite (integration which actually tests the terminal), we can't write any tests and thus no coverage.
exactly. the best time to set up coverage monitoring is *before* you implement the tests
> 
> 1. https://github.com/taiki-e/cargo-llvm-cov is better than tarpualin (due to experience)

i don't know anything about `cargo-llvm-cov`, but I do know that `tarpaulin` has come a *long* way in recent times. Might be worth revisiting.

> 2. We will probably not go with codecov.io and instead I can ask @mitsuhiko to setup coveralls.

i don't really have a preference here. what does coveralls give you that codecov doesn't?

pksunkara:
> i don't know anything about cargo-llvm-cov, but I do know that tarpaulin has come a long way in recent times. Might be worth revisiting.

It's still bad. Especially not running on mac. `llvm-cov` is probably also a better implementation instead of tarpaulin's complicated one.

> what does coveralls give you that codecov doesn't?

Codecov doesn't allow us to compare the changes of coverage in PR. The web app is broken. The only good thing about it is forwarding flags feature. But we don't need really need that in this repo. Thus, coveralls is pretty sufficient.

danieleades:
> Codecov doesn't allow us to compare the changes of coverage in PR

what do you mean by that? Codecov certainly does have this feature. It's the main purpose. Unless I'm misunderstanding?

The only reason the coverage *diff* isn't shown on the report in this PR is because there's no coverage information for the base commit of the diff (it's never been run on master).

Either way, it's not a hill I plan to die on. My bias towards tarpaulin+codecov is that i have a ready-made workflow i was able to copy and paste. If you want to create one using your preferred tools, go for it.

Any coverage is better than none. And the best time to start tracking coverage is *before* you start adding tests.

pksunkara:
Hmm.. The only experience with codecov is at work. And the "Compare" tab has been broken for more than an year and not showing anything. But it looks it's an issue on our side rather than codecov being completely broken for everyone. In that case, I am okay with codecov.

PR #166, opened 2022-01-19
confirm: properly clear the terminal in case of a multi-line prompt string

### LINKED ISSUE #165: Confirmation prompt: non-last line of multiline strings get printed twice
Thanks for this crate, by the way. I could insert a simple terminal prompt without knowing too much about terminals!

**Steps to reproduce.**
Take this small test programm.
```$ cat Cargo.toml
[package]
name = "dialoguer-bug"
version = "0.1.0"
edition = "2021"
[dependencies]
dialoguer = "0.9.0"

$ cat src/main.rs
fn main() {
    let _ = dialoguer::Confirm::new().with_prompt("Prompt string with multiple lines:\nLine two?").interact().map_err(|e| e.to_string());
}
```

1. Build and run the programm: `cargo r`
2. A confirmation prompt appears:
> Prompt string with multiple lines:
> Line two?
3. Answer either yes or no (doesn't matter which).
4. My answer is displayed in the terminal.

**Expected result** The terminal output looks like this:
> Prompt string with multiple lines:
> Line two yes

**Actual result** The prompt is printed twice, resulting in output like this.
> Prompt string with multiple lines:
> Prompt string with multiple lines:
> Line two yes

**Meta.** Tested on both version 0.9 and 0.8 of dialoguer.

grunweg:
Need to clear the current line before clearing previous lines, otherwise a stray final line is left behind. Not sure why.
Fixes #165.

I'm not sure how to add a test. Happy to add one, if somebody can point me at how to do this.
Barring that, I tested this manually, for all four combinations of report={true,false} and wait_for_newline={true,false}.

### DIFF
--- a/src/prompts/confirm.rs
+++ b/src/prompts/confirm.rs
@@ -210,8 +210,10 @@ impl Confirm<'_> {
                         continue;
                     }
                 };
-
+                // `prompt` may be a multi-line string, hence we may need to clear more than the current line.
+                // `clear_last_lines` clears the n lines *before* the current line.
                 term.clear_line()?;
+                term.clear_last_lines(self.prompt.lines().count() - 1)?;
                 render.confirm_prompt(&self.prompt, value)?;
             }
         } else {
@@ -234,7 +236,10 @@ impl Confirm<'_> {
             }
         }
 
+        // `prompt` may be a multi-line string, hence we may need to clear more than the current line.
+        // `clear_last_lines` clears the n lines *before* the current line.
         term.clear_line()?;
+        term.clear_last_lines(self.prompt.lines().count() - 1)?;
         if self.report {
             render.confirm_prompt_selection(&self.prompt, rv)?;
         }


grunweg:
I realised why we need to clear the current line first: clearing the last lines leaves the cursor on the first cleared line --- so to clear the current line first, we'd need to move the cursor down again.

It's easier to clear the current line first, and then clear the last n lines. That way, the cursor will be placed at the last char in the terminal (which is what we want).

src/prompts/confirm.rs
```diff
@@ -210,8 +210,10 @@ impl Confirm<'_> {
                         continue;
                     }
                 };
-
+                // `prompt` may be a multi-line string, hence we may need to clear more than the current line.
+                // `clear_last_lines` clears the n lines *before* the current line.
                 term.clear_line()?;
+                term.clear_last_lines(self.prompt.lines().count() - 1)?;
```
pksunkara, line 216: Can you please convert this change to all the prompt types?

grunweg: Thanks for the feedback! I'm happy to, but I don't see how that makes sense.

I looked at all the other prompts, and and all of them support multi-line prompt strings without issue. This bug only manifests because when trying to clear the current prompt. Confirmation prompts do this, but none of the other prompts do.

More specifically:
- the only other prompt calling the `clear_line` method is `input.rs`; that part of code is not clearing a prompt.
- `clear_last_lines` is used a couple of times, but always to reset the whole terminal (such as `clear_last_lines(paging.capacity)?`).

grunweg: I agree that this logic feels not really specific to confirmation prompts. I'm not sure where to extract it to, though. Can you point me to a better place? Thanks!

pksunkara: Extraction is not what I had in mind. Just copying to other prompts is enough. But as you said, looks like the other prompts don't have this bug. I will need to investigate.

pksunkara: Investigated and looks like yes, bug only happens in this prompt. But instead of doing it like this, can you please replace `term.clear_line()` with `render.clear()` and see if that satisfies for you?

grunweg: I tried; that doesn't fully work. For me, it's easiest to share a test case.
```
fn main() {
    let run = |msg: &str| {
        let _ = dialoguer::Confirm::new().report(true).wait_for_newline(false).with_prompt(msg).interact().map_err(|e| e.to_string());
    };
    run("prompt 1, single line");
    run("prompt 2, single line");
    run("prompt 3, single line");
    run("prompt 4,\n on two lines");
    run("prompt 5,\n on two lines");
    run("prompt 6,\n on \nthree lines");
}
```
Run the code using cargo run. Responds to all prompts with "y" (for example).

Results. Every single-line input is printed twice; multi-line input is not quite correct. More precisely, at the end of the session, I have the following terminal output.
```
Testing no-newline case with reporting
prompt 1, single line [y/n] prompt 1, single line yes
prompt 2, single line [y/n] prompt 2, single line no
prompt 3, single line [y/n] prompt 3, single line no
prompt 4,
 on two lines yesn] 
prompt 5,
 on two lines yesn] 
prompt 6,
 on 
three lines yesn]```
I haven't looked in detail where the bug is; hopefully the above is helpful enough.

grunweg: Oh, and the bug is not specific to the combination of "report=true" and "wait_for_newline=false".

pksunkara: Looks like the bug is with theme renderer. We might be doing different things for confirm prompt rather than the pattern we used in other prompts. (I consider select prompt to be highly polished).

In any case, I would rather appreciate if you could fix it correctly in there. I am not interested in accepting this solution which would just increase the codebase's technical debt. With that said, I understand if you are not willing to continue anymore. In that case, you will have to wait for me to work on 0.7.0 in a few months for fixing this issue

grunweg: Ack, fair enough. It so turned out that I had a spare moment today and looked into fixing this directly -- and the fix was reasonable; I just pushed that. Let me know if you'd like me to change anything.

pksunkara: Hmm.. you sure you pushed it? I don't see it.

grunweg:
Closing in favour of #175.

PR #169, opened 2022-01-26
Fix buffer overflow in completion example

### LINKED ISSUE #168: Buffer overflow panic using completion example
Problematic code: `self.options.iter().filter(|x| s == x[..s.len()])`

This causes a buffer overflow panic whenever the input's length (`s.len()`) is greater than one of the option's string lengths (`x.len()`). To reproduce, type/autocomplete "banana" and then press tab again:

image

bryanhitc:
Resolves [Buffer overflow panic using completion example #168](https://github.com/mitsuhiko/dialoguer/issues/168).

### DIFF
--- a/examples/completion.rs
+++ b/examples/completion.rs
@@ -29,10 +29,9 @@ impl Default for MyCompletion {
 impl Completion for MyCompletion {
     /// Simple completion implementation based on substring
     fn get(&self, input: &str) -> Option<String> {
-        let s = input.to_string();
-        let ss: Vec<&String> = self.options.iter().filter(|x| s == x[..s.len()]).collect();
-        if ss.len() == 1 {
-            Some(ss[0].to_string())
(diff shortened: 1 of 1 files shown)


examples/completion.rs
```diff
@@ -29,10 +29,14 @@ impl Default for MyCompletion {
...
     fn get(&self, input: &str) -> Option<String> {
-        let s = input.to_string();
-        let ss: Vec<&String> = self.options.iter().filter(|x| s == x[..s.len()]).collect();
-        if ss.len() == 1 {
-            Some(ss[0].to_string())
+        let matches = self
+            .options
+            .iter()
+            .filter(|option| option.starts_with(input))
+            .collect::<Vec<_>>();
```
bryanhitc, line 36: This can be optimized to only `.take(2)` since we only care if we have `0`, `1`, or `>= 2` matches, but I don't want to complicate/change the algorithm since it's meant to be a simple example.

Using a trie would be even better, but that should be a separate example/implementation of `Completion` imo.

Issue #163, opened 2022-01-14
Highlight fuzzy select matched indices for the search


craciuncezar:
Thanks for building and maintaining this lib. I have a small suggestion to the `fuzzy_select` prompt:

A nicer touch might be to highlight the matched indices similar to how vscode does it; where chars at matched indices are bold:

![vscodeexample](https://user-images.githubusercontent.com/27342306/149578354-75b9460a-e6f0-4a4c-8dac-6c4ecc56d213.gif)

The `fuzzy-matcher` lib returns the matched indices with the [fuzzy_indices](https://docs.rs/fuzzy-matcher/0.3.7/fuzzy_matcher/trait.FuzzyMatcher.html#tymethod.fuzzy_indices) method.

PR #179, opened 2022-03-06
Feature: Add Fuzzy Select match highlighting

deg0nz:
This implements #163.

In the `SimpleTheme`, the matched characters are printed bold, the default `ColorfulTheme` prints the characters in bold and yellow.

The highlighting is enabled by default and can be turned off via `.highlight_matches(false)` in the prompt builder.

Here's an example:

highlight-fuzzy-indices

### DIFF
--- a/src/prompts/fuzzy_select.rs
+++ b/src/prompts/fuzzy_select.rs
@@ -39,6 +39,7 @@ pub struct FuzzySelect<'a> {
     prompt: String,
     report: bool,
     clear: bool,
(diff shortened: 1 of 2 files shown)


pksunkara (approved):

Araxeus:
There's a feature regression here:
the currently selected option isn't colored anymore

I personally think it looks weird now, could it be added back?

(it's uncolored regardless of `.highlight_matches()`)

PR #183, opened 2022-03-07
Fix: [Fuzzy-Select] Special chars cause panic

### LINKED ISSUE #181: Fuzzy Select: Special characters lead to panic
Entering special characters that consist of more than one char (like ä, ö, ü, é, ê, ...) lead to a panic.

This seems to be due to the fact that we use a range in `search_term[0..cursor_pos]`.

I will check, if [using `Slice`s fixes the issue](https://doc.rust-lang.org/std/primitive.str.html#method.split_at).

fuzzy-select-umlaut

Edit: also found this while implementing #179

deg0nz:
This fixes #181

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ readme = "README.md"
 [features]
 default = ["editor", "password"]
 editor = ["tempfile"]
-fuzzy-select = ["fuzzy-matcher"]
+fuzzy-select = ["fuzzy-matcher", "unicode-segmentation"]
 history = []
 password = ["zeroize"]
 completion = []
@@ -28,6 +28,7 @@ console = "0.15.0"
 tempfile = { version = "3", optional = true }
 zeroize = { version = "1.1.1", optional = true }
 fuzzy-matcher = { version = "0.3.7", optional = true }
+unicode-segmentation = { version = "1.9.0", optional = true }
 
 [[example]]
 name = "password"
--- a/examples/fuzzyselect.rs
+++ b/examples/fuzzyselect.rs
@@ -30,6 +30,9 @@ fn main() {
         "Brown sugar",
         "Blueberry",
         "Burger",
+        "Käse",
+        "Döner",
+        "Blåbär"
     ];
 
     let selection = FuzzySelect::with_theme(&ColorfulTheme::default())
--- a/examples/fuzzyselect.rs
+++ b/examples/fuzzyselect.rs
@@ -32,7 +32,7 @@ fn main() {
         "Burger",
         "Käse",
         "Döner",
-        "Blåbär"
+        "Blåbär",
     ];
 
     let selection = FuzzySelect::with_theme(&ColorfulTheme::default())
--- a/src/prompts/fuzzy_select.rs
+++ b/src/prompts/fuzzy_select.rs
@@ -172,7 +172,10 @@ impl FuzzySelect<'_> {
 
         loop {
             let concatted_search_term = search_term.concat();
-            search_term = concatted_search_term.graphemes(true).map(|s| s.to_string()).collect::<Vec<String>>();
+            search_term = concatted_search_term
+                .graphemes(true)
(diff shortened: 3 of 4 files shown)


deg0nz:
I just realised, that this does not include the changes regarding fuzzy-select highlighting (#179).

I suggest, we wait until a fix for #195 is merged and then adjust this PR here accordingly before merging.

Araxeus:
Now that we can search special characters, I encountered a bug where searching for a left to right char would break the order of the string, after some debugging it seems that chained ANSI escaped chars do not preserve their text order

I've opened an issue in the rust repo, maybe it will be fixed someday
https://github.com/rust-lang/rust/issues/97020

pksunkara:
Ideally, this should be handled by `console`. But  `console` isn't maintained regularly and I don't have time to work on it even though I do have access to it. For that reason, I plan to refactor this crate to use `crossterm`.

deg0nz:
> For that reason, I plan to refactor this crate to use `crossterm`.

Sounds good. My summer is pretty packed already, so I don't much time for coding in my free time. But if this transition is still open in fall/winter, I'm happy to contribute to this.

pksunkara:
I am aiming for summer but realistically I probably can't get to it before fall.

Araxeus:
I'm using `crossterm` in my fork https://github.com/Araxeus/dialoguer
(select and fuzzy-select only)

since it allows me to check for key modifier, which I personally needed

pksunkara:
Nice. Does anyone know how to write tests for crossterm? If we have proper tests for dialoguer, it would make my life so much easier.

Araxeus:
The crossterm source code has a lot of internal tests you can take inspiration from

grunweg:
> I am aiming for summer but realistically I probably can't get to it before fall.

I had some leftover time and started moving to crossterm. Find my partial attempt [here](#213) in case it's helpful.

grunweg:
> The crossterm source code has a lot of internal tests you can take inspiration from

I'd like to add tests verifying the actual output on the terminal: "run this commands on the terminal/spin up this dialogue, then capture the current output", then repeat with more code. I would guess this already exists, but I can't seem to find it.
crossterm's tests are different; I don't see any of this kind. @Araxeus Do you know if such a thing exists?

pksunkara:
Yeah, `crossterm` doesn't really have good TUI based tests.

What I have been looking for recently is a good library to write TUI end-to-end tests. I did a bit of research and found that some python tools use `pyte`(https://pyte.readthedocs.io/en/latest/)

Trying to find something similar in Rust ecosystem pointed me to the following:

* These don't support TUI or colors or inputs
  * https://github.com/assert-rs/assert_cmd
  * https://github.com/assert-rs/trycmd
* These don't support TUI or colors.
  * https://github.com/zhiburt/expectrl
  * https://github.com/philippkeller/rexpect
* These don't support TUI
  *  https://github.com/slowli/term-transcript

For us to have to good testing, we need to either modify `term-transcript` or write a new one on our own based on https://github.com/alacritty/vte.

@epage I would appreciate your thoughts on this since this is related to rust-cli ecosystem as a whole.

epage:
Historically, TUIs have been out of scope of rust-cli (there has been enough to do for normal CLIs) but I expect there is a lot we can improve about the TUI landscape.  atm I don't have a TUI use case to help me dig in and see what the experience is like and speak on it, so any thoughts I have will be theoretical

grunweg:
@psunkara I see; that makes a lot of sense. That's out of my depth. (I'd be happy to help adding tests once there is a testing framework.)

Searching for TUI on crates.io yields the [tui-rs](https://crates.io/crates/tui) and [cursive](https://crates.io/crates/cursive) crates (which are used for building TUI interfaces). I don't know if they allow writing TUI tests; just dropping this in case it's helpful.

Gordon01:
This can be closed as fixed in https://github.com/console-rs/dialoguer/pull/245

PR #190, opened 2022-05-06
Handle selection items overflowing several times

grunweg:
PR #14 was good, but not general enough: there were in fact two different bugs hiding. This PR
- documents the logic of the dix
- fixes an off-by-two error in the logic (which is easy to overlook; I only found that during manual testing),
- fixes clear_preserve_prompt to handle items overflowing several times
- cleans up the code slightly using iterators (happy to drop this commit if you prefer)
See individual commits for details.

### DIFF
--- a/src/prompts/fuzzy_select.rs
+++ b/src/prompts/fuzzy_select.rs
@@ -144,9 +144,9 @@ impl FuzzySelect<'_> {
         let mut sel = self.default;
 
         let mut size_vec = Vec::new();
-        for items in self.items.iter().as_slice() {
-            let size = &items.len();
-            size_vec.push(*size);
+        for item in self.items.iter().as_slice() {
+            // Formatting each item adds two more characters.
+            size_vec.push(item.len() + 2);
         }
 
         // Fuzzy matcher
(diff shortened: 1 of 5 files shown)


grunweg:
There's a deeper issue underlying this: for any output line which overflows the terminal (i.e. is longer than the terminal width), we only increase the height by one --- even if the line can span several lines on the terminal. When clearing the terminal, we only take the height into account, i.e. will not clear enough output.
This bug also exists is TermThemeRender::clear.

A deeper fix would be to record the length of each line of output, and take this into account when clearing the terminal. (I have prototyped this; it seems workable, if slightly fiddly.)

Counter-arguments I can think of
1. Not every terminal has a size (see console::TermTarget::ReadWritePair).
Solution: when no size is available, just use the default size. That's no worse than the current state.

2. This doesn't handle terminal resizing: if we just check the size at clearing time, that can result in both over- or underclearing, depending on whether the terminal was resized in the meantime.
Solution: don't record the number of additional lines, but the lengths of inputs. Compute the height to clear when clearing. Is not hard; we just store some additional data for each line we wrote.

3. Do we need this? This issue is particularly pronounced for selection prompts (as the selection items need to be drawn over and over, potentially). It's just not as bad for the other prompts.

@psunkara (or other maintainers). Would you prefer to see a "more right" fix instead?

grunweg:
Looks like it.

PR #191, opened 2022-05-06
fix: also reset self.prompt_height in `TermThemeRenderer::clear`.

grunweg:
This appears like an oversight: we clear the last `self.height+self.prompt_height` lines from the output; we set `height` to 0, but not `prompt_height`. We should also reset prompt_height.

This can cause issues if we were to call clear() twice in a row: we would clear prompt_height lines more than we should.

### DIFF
--- a/src/theme.rs
+++ b/src/theme.rs
@@ -843,6 +843,7 @@ impl<'a> TermThemeRenderer<'a> {
         self.term
             .clear_last_lines(self.height + self.prompt_height)?;
         self.height = 0;
+        self.prompt_height = 0;
         Ok(())
     }
 


grunweg:
There's a simple "manual test case": apply the following patch on top of the change and execute any confirmation prompt.
```
diff --git a/src/prompts/confirm.rs b/src/prompts/confirm.rs
--- a/src/prompts/confirm.rs
+++ b/src/prompts/confirm.rs
@@ -237,6 +237,10 @@ impl Confirm<'_> {
         term.clear_line()?;
         if self.report {
             render.confirm_prompt_selection(&self.prompt, rv)?;
+            render.clear()?;
+            render.clear()?;
+            render.confirm_prompt_selection(&self.prompt, rv)?;
+
         }
         term.show_cursor()?;
         term.flush()?;
diff --git a/src/theme.rs b/src/theme.rs
--- a/src/theme.rs
+++ b/src/theme.rs
@@ -843,7 +843,7 @@ impl<'a> TermThemeRenderer<'a> {
         self.term
             .clear_last_lines(self.height + self.prompt_height)?;
         self.height = 0;
-        self.prompt_height = 0;
+        //self.prompt_height = 0;
         Ok(())
     }
 ```
Run a simple prompt, e.g. the following (in a crate with dialoguer as dependency)
```
fn main() -> std::io::Result<()> {
dialoguer::Confirm::new().with_prompt("prompt string").interact()?;
}
```
Observe. The prompt string and lines before it are cleared. Making the prompt string longer exerbates this.

PR #199, opened 2022-05-14
Fix: [Fuzzy-Select] Apply active style to active item

### LINKED ISSUE #195: Fuzzy Select doesn't color currently selected item
I think match highlighting in https://github.com/mitsuhiko/dialoguer/pull/179 caused an undocumented feature regression/bug?

Fuzzy Select doesn't color the currently selected item

in `v0.10.0` currently selected item was highlighted in blue (using `ColorfulTheme`)
but now in `v0.10.1` it isn't anymore...

It's important to note that it doesn't work regardless of if `.highlight_matches()` is `false` or `true` 

(since `.highlight_matches(false)` should behave exactly like in `v0.10.0`, I think this is a bug)

----
On a side note:

I set the highlight style to be Bold only using:
```rust
let theme = ColorfulTheme {
    fuzzy_match_highlight_style: console::Style::new().for_stderr().bold(),
    ..Default::default()
};
```

I think this looks better, so maybe `bold` should be the default instead of `yellow`+`bold`
(especially since the yellow color clash with the currently selected blue color that hopefully will be back)

----

deg0nz:
This fixes #195.

I decided to go for the bold-only option (No3 in https://github.com/mitsuhiko/dialoguer/issues/195#issuecomment-1121610065).

### DIFF
--- a/src/theme.rs
+++ b/src/theme.rs
@@ -340,7 +340,7 @@ impl Default for ColorfulTheme {
             #[cfg(feature = "fuzzy-select")]
             fuzzy_cursor_style: Style::new().for_stderr().black().on_white(),
             #[cfg(feature = "fuzzy-select")]
-            fuzzy_match_highlight_style: Style::new().for_stderr().bold().yellow(),
+            fuzzy_match_highlight_style: Style::new().for_stderr().bold(),
             inline_selections: true,
         }
     }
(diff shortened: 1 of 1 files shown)


mmun:
Looks good to me 👍

pksunkara (approved):

PR #208, opened 2022-07-09
Home, End, Del operations for Input struct prompt

alexzanderr:
Hello again.

In this PR I've added
- `Home, End, Del operations` for `Input` struct prompt.
- an empty `rustfmt.toml` to specify that rustfmt should format with default settings (why is that? because I have a `~/.config/rustfmt/rustfmt.toml` on my system)
- fixes to the source code because clippy wasnt very happy.

I've run:
- `cargo build --all-features`
- `cargo clippy --all-features`
- `cargo test --lib`
- `cargo fmt -- --check`

all successes.

I've tested the feature from my cli and worked on most (almost all) cases, so I think would work.

Cant wait for response!

Thanks in advance.

### DIFF
--- a/src/prompts/input.rs
+++ b/src/prompts/input.rs
@@ -443,7 +443,6 @@ where
                 self.initial_text = Some(String::from(""));
             }
 
-
             let input = chars.iter().collect::<String>();
 
             term.clear_line()?;
--- a/src/prompts/input.rs
+++ b/src/prompts/input.rs
@@ -435,10 +435,6 @@ where
                     _ => (),
                 }
             }
-            // when you got initial text already
-            // and you deleted all the chars with backspace or del
-            // and then you press enter, the initial text will reapear
-            // so the initial modified to "" to not appear anymore
             if self.initial_text.is_some() && chars.is_empty() {
                 self.initial_text = Some(String::from(""));
             }
--- a/src/prompts/input.rs
+++ b/src/prompts/input.rs
@@ -309,7 +309,7 @@ where
                     }
                     // press Del key to delete the char in front of the cursor
                     // cursor stays in the same position
-                    Key::Del if 0 <= position && position < chars.len() => {
+                    Key::Del if position < chars.len() => {
                         // first get the tail before deletion
                         let tail = chars[position + 1..].iter().collect::<String>();
                         // move visually 1 unit to right
@@ -317,6 +317,7 @@ where
                         // delete all the chars from right to the pos of cursor (which is to left side)
                         term.clear_chars(1)?;
                         chars.remove(position);
+
                         // append visually the tail to the user written text
                         term.write_str(&tail)?;
                         // move visually to the left length of tail; why ?
@@ -434,6 +435,15 @@ where
                     _ => (),
                 }
             }
+            // when you got initial text already
+            // and you deleted all the chars with backspace or del
+            // and then you press enter, the initial text will reapear
+            // so the initial modified to "" to not appear anymore
+            if self.initial_text.is_some() && chars.is_empty() {
+                self.initial_text = Some(String::from(""));
+            }
+
+
             let input = chars.iter().collect::<String>();
 
             term.clear_line()?;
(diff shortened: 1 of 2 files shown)


src/prompts/input.rs
```diff
@@ -286,10 +286,46 @@ where
                 term.write_str(initial)?;
                 chars = initial.chars().collect();
                 position = chars.len();
+                term.flush()?;
```
alexzanderr, line 289: here refresh was missing and the initial text wouldnt appear after cold run

src/prompts/input.rs
```diff
@@ -399,6 +435,14 @@ where
                     _ => (),
                 }
             }
+            // when you got initial text already
+            // and you deleted all the chars with backspace or del
+            // and then you press enter, the initial text will reapear
+            // so the initial modified to "" to not appear anymore
+            if self.initial_text.is_some() && chars.is_empty() {
```
alexzanderr, line 442: when you got initial text already
and you deleted all the chars with backspace or del
and then you press enter, the initial text will reapear
so the initial modified to "" to not appear anymore

src/theme.rs
```diff
@@ -222,7 +222,7 @@ pub trait Theme {
         write!(f, "{} ", if active { ">" } else { " " })?;
 
         if highlight_matches {
-            if let Some((_score, indices)) = matcher.fuzzy_indices(text, &search_term) {
+            if let Some((_score, indices)) = matcher.fuzzy_indices(text, search_term) {
```
alexzanderr, line 225: search_term is already `&str` so `&search_term` is redundant

src/theme.rs
```diff
@@ -258,7 +258,7 @@ pub trait Theme {
             write!(f, "{}{}{}", st_head, st_cursor, st_tail)
         } else {
             let cursor = "|".to_string();
-            write!(f, "{}{}", search_term.to_string(), cursor)
```
alexzanderr, line 261: unncessary convertion to String

src/theme.rs
```diff
@@ -649,12 +649,12 @@ impl Theme for ColorfulTheme {
                         } else {
                             write!(f, "{}", self.fuzzy_match_highlight_style.apply_to(c))?;
                         }
+                    } else if active {
+                        // since in lib.rs is deny clippy::all
+                        // the old else if was converted to this one
+                        write!(f, "{}", self.active_item_style.apply_to(c))?;
                     } else {
```
alexzanderr, line 656: unnecessary else block

src/theme.rs
```diff
@@ -696,13 +696,7 @@ impl Theme for ColorfulTheme {
...
         } else {
             let cursor = self.fuzzy_cursor_style.apply_to(" ");
-            write!(
-                f,
-                "{} {}{}",
-                &self.prompt_suffix,
-                search_term.to_string(),
-                cursor
-            )
+            write!(f, "{} {}{}", &self.prompt_suffix, search_term, cursor)
```
alexzanderr, line 699: this formatted using `cargo fmt --all`, the default settings, as you wanted.

src/prompts/input.rs
```diff
@@ -399,6 +435,10 @@ where
                     _ => (),
                 }
             }
+            if self.initial_text.is_some() && chars.is_empty() {
```
alexzanderr, line 438: when you got initial text already
and you deleted all the chars with backspace or del
and then you press enter, the initial text will reapear
so the initial modified to "" to not appear anymore

src/prompts/input.rs
alexzanderr: Just found out that running this example with `buffered stdout`, `term.flush()?;` is necessary
```rust
    let term = Term::buffered_stdout();
    let input: String = Input::with_theme(&ColorfulTheme::default())
        .with_prompt("Your name")
        .with_initial_text("abcdef")
        .allow_empty(true)
        .interact_on(&term)?;
```

but when using just stdout
```rust
    let term = Term::stdout();
```
the refresh `term.flush()?;` is not needed anymore.

what is going on ?

PR #213, opened 2022-08-18
WIP: move from console to crossterm

grunweg:
I've started moving this crate to crossterm [as hinted here](https://github.com/mitsuhiko/dialoguer/pull/183#issuecomment-1140419654) and figured I'd share my progress.

**Scope of task.** TTBOMK, there are three parts to this migration, corresponding to the different console types used:
- replace `console::Key` by `crossterm::KeyCode`
- replace `console::Term` by a `crossterm` equivalent
- replace `console::{style, Style, StyledObject}` by crossterm equivalents

The first item is tedious, but pretty straightforward; all necessary changes are mechanical. Replacing console::Term is more subtle; I've introduced bugs in the porting process more than once (which manual testing has caught). I haven't looked deeply into the third part yet. console's and crossterm's machinery around terminal styling are somewhat different; I haven't wrapped my head around that. I'd leave that for another PR.

**Testing.** I mostly used manual testing to verify my changes.
- `cargo test`, `cargo clippy` and `cargo fmt` pass
- run all the examples and verify their output looks sane: `cargo run --example confirm` and likewise

**Status.**
- [x] replace `console::Key` by `crossterm::KeyCode`
- [] replace `console::Term` by a `crossterm` equivalent
  - [x] confirmation prompt: compiles and tested manually; works
  - [x] input prompt: compiles; manual tests reveal some regressions
  - [x] fuzzy_select: compiles; manual tests reveal some regressions
  - [x] sort, select, multiselect: compiles; untested so far
  - [] password: mostly ported, except for read_line_secure (and untested)
  - [] `buffered.rs` example: uses `crossterm::Term::buffered_stderr`
- [] replace `console::{style, Style, StyledObject}` by crossterm equivalents: not started

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ console = "0.15.0"
 tempfile = { version = "3", optional = true }
 zeroize = { version = "1.1.1", optional = true }
 fuzzy-matcher = { version = "0.3.7", optional = true }
+crossterm = "0.24.0"
 
 [[example]]
 name = "password"
--- a/examples/buffered.rs
+++ b/examples/buffered.rs
@@ -1,4 +1,4 @@
-use console::Term;
+fn main() {} /*use console::Term;
 use dialoguer::{theme::ColorfulTheme, Confirm, Input, MultiSelect, Select, Sort};
 
 fn main() {
@@ -40,3 +40,4 @@ fn main() {
         .interact_on(&term)
         .unwrap();
 }
+*/
--- a/examples/buffered.rs
+++ b/examples/buffered.rs
@@ -1,4 +1,4 @@
-fn main() {} /*use console::Term;
+fn main() {} // TODO fix this! /*use console::Term;
 use dialoguer::{theme::ColorfulTheme, Confirm, Input, MultiSelect, Select, Sort};
 
 fn main() {
--- a/examples/buffered.rs
+++ b/examples/buffered.rs
@@ -1,4 +1,5 @@
-fn main() {} // TODO fix this! /*use console::Term;
+fn main() {} // TODO fix this!
+/*use console::Term;
 use dialoguer::{theme::ColorfulTheme, Confirm, Input, MultiSelect, Select, Sort};
 
 fn main() {
--- a/examples/completion.rs
+++ b/examples/completion.rs
@@ -1,4 +1,4 @@
-use dialoguer::{theme::ColorfulTheme, Completion, Input};
+fn main() {} /* use dialoguer::{theme::ColorfulTheme, Completion, Input};
 
 fn main() -> Result<(), std::io::Error> {
     println!("Use the Right arrow or Tab to complete your command");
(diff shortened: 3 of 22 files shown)


grunweg:
I have solved the borrowing issue above; with the latest push, all prompts compile now. (The one exception is the buffered example, which I haven't ported. cargo fmt, cargo clippy and cargo test also pass.) I haven't tested the prompts any further.

pksunkara:
FWIW, this PR is not going to go to waste. We just need to implement the testing framework is all.

pksunkara:
I have been making some good progress in figuring out how to do unit testing for TUIs. I should have something soon.

grunweg:
@pksunkara This sounds exciting! Any news since March?

(I've seen there are merge conflicts on this PR. I'll resolve them when there is a clearer story forward - which, as I understand it, is blocked on figuring out a better unit testing story.)

Gordon01:
I would like the part with switching to `crossterm` to be merged. 
`console` has bugs like https://github.com/console-rs/console/pull/182 which affect `dialoguer`.

Hopefully it can make Windows experience better.

pksunkara:
Agreed. I have some work locally that I will look to push. @Gordon01 if you are interested, you can also take charge of it later.

mitsuhiko:
I do not want to do this merge without considering what this means. There are a lot of projects using console and dialoguer in one way or another and this switch would have to be considered properly first. In general all of these crates (dialoguer included) are quite buggy as a hole and have frequent regressions, I'm not convinced that this move will significantly improve on this.

mitsuhiko:
To make this more actionable: I rather see a fork of this library out of this organization to be based on crossterm than to make such a major departure here.

Gordon01:
> I'm not convinced that this move will significantly improve on this.

Agree. Wasn't familiar with the organization of the project but now I see.

> In general all of these crates (dialoguer included) are quite buggy as a hole and have frequent regressions

Generally, they work well. But definitely needs a few touches. Especially for Windows, ha-ha.

> if you are interested, you can also take charge of it later.

@pksunkara great!

Thank you for your effort and quick merge of my patches. 
I recently started to use `dialoguer` for my CLI to make menus like "Select a group for user". Makes the text UIs easier to use. Great that it supports passwords too.

image | image
-- | --

Gordon01:
@pksunkara I believe, this PR may be closed now as this crate would stay on `console`.

It's also listed as a 0.11 milestone which is definitely not true now)

LingMan:
Before anyone goes ahead and forks `dialoguer`, I'd like to point out that `inquire` already uses `crossterm` by default but also supports `console` and `termion` as alternative backends.
Might make more sense to contribute there instead of creating yet another prompting library.

Gordon01:
I definitely wouldn't do it, since we now fixed the most annoying bugs of non-ascii characters support.

@pksunkara can this be closed now?

PR #214, opened 2022-08-24
Enable customization of number of rows for FuzzySelect

firebolt55439:
Hi!

In this PR, I've made a small but useful extension to `FuzzySelect` that enables the user to decide how many options will be displayed at one time, rather than the default of the terminal height minus 2.

Example code:
```rust
FuzzySelect::with_theme(&ColorfulTheme::default())
                .with_prompt("What is your guess?")
                .with_visible_term_rows(5_usize)
                .items(word_slice)
                .report(true)
                .default(0)
                .interact()?;
```

Here's an example of how it looks with `visible_term_rows = 5`:

image

I'm still somewhat new to Rust, let me know if there's any code style that can be improved.

### DIFF
--- a/src/prompts/fuzzy_select.rs
+++ b/src/prompts/fuzzy_select.rs
@@ -172,7 +172,10 @@ impl FuzzySelect<'_> {
 
         // Subtract -2 because we need space to render the prompt.
         let visible_term_rows = (term.size().0 as usize).max(3) - 2;
-        let visible_term_rows = self.visible_term_rows.unwrap_or(visible_term_rows);
+        let visible_term_rows = self
+            .visible_term_rows
+            .unwrap_or(visible_term_rows)
+            .min(visible_term_rows);
         // Variable used to determine if we need to scroll through the list.
         let mut starting_row = 0;
 
--- a/src/prompts/fuzzy_select.rs
+++ b/src/prompts/fuzzy_select.rs
@@ -40,7 +40,7 @@ pub struct FuzzySelect<'a> {
     report: bool,
     clear: bool,
     highlight_matches: bool,
-    visible_term_rows: Option<usize>,
+    max_length: Option<usize>,
     theme: &'a dyn Theme,
 }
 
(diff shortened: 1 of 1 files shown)


src/prompts/fuzzy_select.rs
```diff
@@ -163,6 +172,7 @@ impl FuzzySelect<'_> {
 
         // Subtract -2 because we need space to render the prompt.
         let visible_term_rows = (term.size().0 as usize).max(3) - 2;
+        let visible_term_rows = self.visible_term_rows.unwrap_or(visible_term_rows);
```
pksunkara, line 175: Let's do a min here.

firebolt55439: Fixed.

firebolt55439:
@pksunkara I've addressed your feedback, mind giving this another glance?

src/prompts/fuzzy_select.rs
```diff
@@ -110,6 +111,14 @@ impl FuzzySelect<'_> {
         self
     }
 
+    /// Sets the number of visible options.
```
pksunkara, line 114:
```suggestion
    /// Sets the maximum number of visible options.
```

src/prompts/fuzzy_select.rs
```diff
@@ -110,6 +111,14 @@ impl FuzzySelect<'_> {
...
+    ///
+    /// The default is the height of the terminal minus 2.
+    pub fn with_visible_term_rows(&mut self, rows: usize) -> &mut Self {
```
pksunkara, line 117: Let's name the fn and the member as `max_length` to be consistent.

pksunkara (changes requested):
Sorry for the extra roundtrip, I should have commented on this earlier.

src/prompts/fuzzy_select.rs
firebolt55439: Fixed.

firebolt55439:
Updated according to your feedback @pksunkara .

pksunkara (approved):

PR #215, opened 2022-08-29
feat(fuzzy): implement multi-fuzzy-selection

RobWalt:
This commit includes a working example to showcase the feature.

Items are selected with spacebar and the selection is confirmed with enter.

---

Short background story:

I've been using the command line tool `glab` to work with my gitlab repos. There, some subcommands use a fuzzy-selection-search to select e.g. labels of issues. I found this feature very handy and wanted to use it for shell-scripts etc, but couldn't find a CLI tool or library that provides something like that. I hope you'll like it as much as I do!

To be fair, I have to mention, that the implementation of this feature has one flaw: You can't specifically search for some search term which includes spaces. But in practice that's almost never the case and in the few cases where you search terms with spaces, you can just skip them while typing in your search term and the fuzzy match algo will work great anyways.

I'm ready to make changes, if you request some.

### DIFF
--- a/src/prompts/multi_fuzzy_select.rs
+++ b/src/prompts/multi_fuzzy_select.rs
@@ -299,9 +299,12 @@ impl MultiFuzzySelect<'_> {
                     term.flush()?;
                 }
                 Key::Char(' ') => {
-                    let sel_string = filtered_list[sel].1;
-                    if let Some(sel_string_pos_in_items) =
-                        self.items.iter().position(|item| item.eq(sel_string))
+                    if let Some(sel_string_pos_in_items) = filtered_list
+                        .get(sel)
+                        .map(|selection| selection.1)
+                        .and_then(|sel_string| {
+                            self.items.iter().position(|item| item.eq(sel_string))
+                        })
                     {
                         checked[sel_string_pos_in_items] = !checked[sel_string_pos_in_items];
                     }
--- a/src/prompts/multi_fuzzy_select.rs
+++ b/src/prompts/multi_fuzzy_select.rs
@@ -185,7 +185,7 @@ impl MultiFuzzySelect<'_> {
 
         loop {
             render.clear()?;
-            render.fuzzy_select_prompt(self.prompt.as_str(), &search_term, position)?;
+            render.multi_fuzzy_select_prompt(self.prompt.as_str(), &search_term, position)?;
 
             // Maps all items to a tuple of item and its match score.
             let mut filtered_list = self
--- a/src/prompts/multi_fuzzy_select.rs
+++ b/src/prompts/multi_fuzzy_select.rs
@@ -277,7 +277,7 @@ impl MultiFuzzySelect<'_> {
                     let selected_items = checked
                         .into_iter()
                         .enumerate()
-                        .filter_map(|(idx, checked)| checked.then_some(idx))
+                        .filter_map(|(idx, checked)| checked.then(|| idx))
                         .collect::<Vec<_>>();
 
                     term.show_cursor()?;
--- a/src/prompts/multi_fuzzy_select.rs
+++ b/src/prompts/multi_fuzzy_select.rs
@@ -43,6 +43,7 @@ pub struct MultiFuzzySelect<'a> {
     report: bool,
     clear: bool,
     highlight_matches: bool,
+    max_length: Option<usize>,
     theme: &'a dyn Theme,
 }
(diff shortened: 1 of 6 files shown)


src/theme.rs
```diff
@@ -238,6 +238,43 @@ pub trait Theme {
         write!(f, "{}", text)
     }
 
+    /// Formats a fuzzy multi select prompt item.
+    #[cfg(feature = "fuzzy-select")]
+    fn format_fuzzy_multi_select_prompt_item(
```
pksunkara, line 243:
```suggestion
    fn format_multi_fuzzy_select_prompt_item(
```

src/theme.rs
```diff
@@ -884,6 +921,29 @@ impl<'a> TermThemeRenderer<'a> {
         })
     }
 
+    #[cfg(feature = "fuzzy-select")]
+    pub fn fuzzy_select_multi_prompt_item(
```
pksunkara, line 925:
```suggestion
    pub fn multi_fuzzy_select_prompt_item(
```

src/prompts/multi_fuzzy_select.rs
```diff
@@ -0,0 +1,334 @@
...
+        // Variable used to determine if we need to scroll through the list.
+        let mut starting_row = 0;
+
+        term.hide_cursor()?;
+
+        let mut checked: Vec<bool> = self.defaults.clone();
+
+        loop {
+            render.clear()?;
+            render.fuzzy_select_prompt(self.prompt.as_str(), &search_term, position)?;
```
pksunkara, line 191: Multi select prompt

src/prompts/multi_fuzzy_select.rs
```diff
@@ -0,0 +1,334 @@
...
+        self._interact_on(term, false)
+    }
+
+    /// Like `interact` but allows a specific terminal to be set.
+    #[inline]
+    pub fn interact_on_opt(&self, term: &Term) -> io::Result<Vec<usize>> {
+        self._interact_on(term, true)
+    }
+
+    /// Like `interact` but allows a specific terminal to be set.
```
pksunkara, line 163: No doc for this since it's not `pub`.

Please fix mistake in `fuzzy_select.rs` too.

src/prompts/multi_fuzzy_select.rs
```diff
@@ -0,0 +1,334 @@
...
+    #[inline]
+    pub fn interact(&self) -> io::Result<Vec<usize>> {
+        self.interact_on(&Term::stderr())
+    }
+
+    /// Enables user interaction and returns the result.
+    ///
+    /// The user can select the items using 'Enter' and the index of selected item will be returned.
+    /// The dialog is rendered on stderr.
+    /// Result contains `Some(index)` if user hit 'Enter' or `None` if user cancelled with 'Esc' or 'q'.
```
pksunkara, line 145: Above 2 statements need to be updated.

src/prompts/multi_fuzzy_select.rs
```diff
@@ -0,0 +1,334 @@
...
+        self.interact_on_opt(&Term::stderr())
+    }
+
+    /// Like `interact` but allows a specific terminal to be set.
+    #[inline]
+    pub fn interact_on(&self, term: &Term) -> io::Result<Vec<usize>> {
+        self._interact_on(term, false)
+    }
+
+    /// Like `interact` but allows a specific terminal to be set.
```
pksunkara, line 157: `interact_opt`

Would appreciate if you can fix this mistake in `fuzzy_select.rs` too.

pksunkara (changes requested):
Also missing `format_multi_fuzzy_select_prompt_item` and `format_multi_fuzzy_select_prompt` in `ColorfulTheme`

src/prompts/multi_fuzzy_select.rs
RobWalt: I think this can stay maybe? It only renders the line that shows the current fuzzy search term (if I'm not mistaken) and this isn't really different in the multi-fuzzy prompt.

I could still copy the functionality to separate concerns. What do you think here?

RobWalt: I just added it for now together with the missing functions on `ColorfulTheme`.

RobWalt:
Fixed everything so far, thanks for the fast an thorough review 👍🏼

PR #219, opened 2022-09-08
Add `validate_with` function to password prompt

### LINKED ISSUE #150: Recommend adding `validate_with` function to password prompt
In some cases, we need to validate the password whether it is not strong enough, like `123456` or it contains the user's birthday number, or we require the password to be longer than a number.

qt2:
Closes #150 

Implement validation function in `Password` which would be useful for checking the strength of passwords. This PR includes:

- Adding `validator` property and `validate_with` method to `Password`
- Refactoring the password confirmation process in order to reduce complexity
- Adding validation process in `interact_on` method
- Minor update on an example code

### DIFF
--- a/src/prompts/password.rs
+++ b/src/prompts/password.rs
@@ -8,7 +8,7 @@ use crate::{
 use console::Term;
 use zeroize::Zeroizing;
 
-type PasswordValidatorCallback<'a> = Box<dyn Fn(&str) -> Option<String> + 'a>;
+type PasswordValidatorCallback<'a> = Box<dyn Fn(&String) -> Option<String> + 'a>;
 
 /// Renders a password input prompt.
 ///
@@ -87,7 +87,7 @@ impl<'a> Password<'a> {
     /// # use dialoguer::Password;
     /// let password: String = Password::new()
     ///     .with_prompt("Enter password")
-    ///     .validate_with(|input: &str| -> Result<(), &str> {
+    ///     .validate_with(|input: &String| -> Result<(), &str> {
     ///         if input.len() > 8 {
     ///             Ok(())
     ///         } else {
@@ -104,7 +104,7 @@ impl<'a> Password<'a> {
     {
         let old_validator_func = self.validator.take();
 
-        self.validator = Some(Box::new(move |value: &str| -> Option<String> {
+        self.validator = Some(Box::new(move |value: &String| -> Option<String> {
             if let Some(old) = &old_validator_func {
                 if let Some(err) = old(value) {
                     return Some(err);
--- a/src/prompts/password.rs
+++ b/src/prompts/password.rs
@@ -8,6 +8,8 @@ use crate::{
 use console::Term;
 use zeroize::Zeroizing;
 
+type PasswordValidatorCallback<'a> = Box<dyn FnMut(&String) -> Option<String> + 'a>;
+
 /// Renders a password input prompt.
 ///
 /// ## Example usage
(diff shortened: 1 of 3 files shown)


smbpndk:
Are you still working on this? I check the test it's failing

src/prompts/password.rs
```diff
@@ -73,44 +77,87 @@ impl Password<'_> {
...
     /// If the user confirms the result is `true`, `false` otherwise.
     /// The dialog is rendered on stderr.
-    pub fn interact(&self) -> io::Result<String> {
+    pub fn interact(&mut self) -> io::Result<String> {
         self.interact_on(&Term::stderr())
     }
 
     /// Like `interact` but allows a specific terminal to be set.
-    pub fn interact_on(&self, term: &Term) -> io::Result<String> {
+    pub fn interact_on(&mut self, term: &Term) -> io::Result<String> {
```
pksunkara, line 130: Do we need mut here? That's the only concern for this PR.

qt2:
> Are you still working on this? I check the test it's failing

Sorry for the late reply. The test failure was due to an incorrectly configured import in doc-test. I will fix it and commit it later. Thank you for pointing this out.

src/prompts/password.rs
qt2: Mut was set because `Input::interact` and `Input::interact_on` have mut, and their code was ported. Since it did not seem to be much of a problem in practical use, I used it as is. However, I will consider modifying it if necessary.

pksunkara: That's because of different reasons IIRC.

qt2: I see. I've just updated the location you mentioned and hope it solves the issue.

smbpndk (approved):
LGTM

Tested on [my project](https://github.com/smbpndk/smbpndk-cli/blob/feature/password-validator/src/main.rs#L135).

src/prompts/password.rs
```diff
@@ -73,6 +76,47 @@ impl Password<'_> {
...
+    ///             Err("Password must be longer than 8")
+    ///         }
+    ///     })
+    ///     .interact()
+    ///     .unwrap();
+    /// ```
+    pub fn validate_with<V, E>(&mut self, validator: V) -> &mut Self
+    where
+        V: Fn(&String) -> Result<(), E> + 'a,
+        E: ToString,
```
pksunkara, line 100:
```suggestion
        V::Err: ToString,
```

pksunkara (commented):
Just one small change left.

qt2:
@pksunkara 

In order for `interact_on` to function without mut, the validator needs `Fn` instead of `FnMut`, so a new `PasswordValidator` trait was added separately from the `Validator` for Input.  Also replaced `&String` with `&str` where possible.

pksunkara:
You don't need a password validator trait and the String -> str. I won't be merging them.

qt2:
@pksunkara 

Hmm, I have no idea using `V::Err` in `validate_with` method.  Simply adopting your change suggestion raised an error "associated type `Err` not found for `V`".  Could you tell me some workaround on it, if possible?

qt2:
If using `E` instead of `V::Err` is fine, then I'll revert the last commit.

pksunkara:
I was mistaken, `Input` has a trait already. So, this trait is okay with me.

The only change I would request is to keep the original `&String` and not `&str`. The validate function should be dealing in `&T` where T is the output of the prompt, which is `String` in the password prompt.

pksunkara (approved):

smbpndk:
Hey @pksunkara, how's the release process for this?

pksunkara:
Done.

PR #221, opened 2022-10-14
Adds the `full` feature convention for convenience

pflakus:
I added a mention to the available features in the main documentation (since I needed to look in the Cargo.toml for them myself :sweat_smile: )  
In the second commit I added the `full` feature for convenience, because that seems to be a convention across popular crates. I'm happy to drop that commit and just keep the first one, if you prefer not to add that feature. :)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,6 +17,17 @@
 //! * Fuzzy select prompt
 //! * Other kind of prompts
 //! * Editor launching
+//!
+//! # Crate Features
+//!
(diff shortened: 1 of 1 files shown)


pksunkara:
Thanks for the docs. I would prefer if the second  commit is dropped. Thanks.

pflakus:
Sure thing. I just resubmitted the branch with just the first commit. Let me know if there's anything else to review.

pksunkara (approved):

PR #222, opened 2022-10-17
Making default color scheme compatible for windows terminal displays

robinbernon:
Fixing default color scheme to be compatible with windows terminal displays.

### DIFF
--- a/src/theme.rs
+++ b/src/theme.rs
@@ -318,24 +318,39 @@ pub struct ColorfulTheme {
 
 impl Default for ColorfulTheme {
     fn default() -> ColorfulTheme {
+        #[cfg(not(target_os = "windows"))]
+        mod constants
+        {
+            pub(super) const heavy_check_mark: &str = "✔";
+            pub(super) const heavy_ballot_x: &str = "✘";
+            pub(super) const heavy_right_pointing_angle_quotation_mark_ornament: &str = "❯";
+        }
+        #[cfg(target_os = "windows")]
+        mod constants
+        {
+            pub(super) const heavy_check_mark: &str = "√"; // (subsitute char)
+            pub(super) const heavy_ballot_x: &str = "×"; // (subsitute char)
+            pub(super) const heavy_right_pointing_angle_quotation_mark_ornament: &str = ">"; // (subsitute char)
+        }
+
         ColorfulTheme {
             defaults_style: Style::new().for_stderr().cyan(),
             prompt_style: Style::new().for_stderr().bold(),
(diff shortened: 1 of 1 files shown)


robinbernon:
@pksunkara thoughts?

pksunkara:
I am thinking that this should not be gated on OS but on term application's unicode support.

robinbernon:
These Unicode characters are just not supported on any windows terminals right now. Perhaps a request to support them could be filed somewhere but for now I think the changes will have to be made here.

robinbernon:
> I am thinking that this should not be gated on OS but on term application's unicode support.

I'm guessing this is just a windows problem. Guessing all Unix based systems should be able to support the same list of Unicode chars?

pksunkara:
Please use https://docs.rs/console/latest/console/struct.TermFeatures.html#method.wants_emoji instead of gating on OS. And please make sure to test it manually on your terminal.

robinbernon:
> Please use https://docs.rs/console/latest/console/struct.TermFeatures.html#method.wants_emoji instead of gating on OS. And please make sure to test it manually on your terminal.

@pksunkara I've done as asked. It's decided by .wants_emoji() now :)

src/theme.rs
```diff
@@ -318,30 +318,58 @@ pub struct ColorfulTheme {
...
-            checked_item_prefix: style("✔".to_string()).for_stderr().green(),
-            unchecked_item_prefix: style("✔".to_string()).for_stderr().black(),
-            picked_item_prefix: style("❯".to_string()).for_stderr().green(),
-            unpicked_item_prefix: style(" ".to_string()).for_stderr(),
-            #[cfg(feature = "fuzzy-select")]
-            fuzzy_cursor_style: Style::new().for_stderr().black().on_white(),
-            #[cfg(feature = "fuzzy-select")]
-            fuzzy_match_highlight_style: Style::new().for_stderr().bold(),
-            inline_selections: true,
+        if Term::stdout().features().wants_emoji() {
```
pksunkara, line 321: You can store this in a variable and use it in emoji lines instead of this big if condition block.

robinbernon: Sorry do you mind clarifying what you mean here? I dont understand. Do you want a variable for the term features?

pksunkara:
```rs
let has_emoji = Term::stderr().features().wants_emoji();

// later
ColorfulTheme {
  error_prefix: style((if has_emoji { "✘" } else { "×" }).to_string()).for_stderr().red(),
}
```

Gordon01:
Hey, @robinbernon can you make the changes, please? It would be valuable for Windows users.

PR #223, opened 2022-10-22
Rewrite several prompts to take `IntoIterator<Item: ToString>` instead of `&[T]`

waterlubber:
This allows for much more flexibility when generating prompts. The user can pass in, for example, the output of a `.map` without having to collect it into a vector and provide a slice.
Since all these functions do is run a simple for loop over the input, it's an easy change to make. Existing code should continue to work, as &[T] implements IntoIterator<Item: T> by default.

### DIFF
--- a/src/prompts/fuzzy_select.rs
+++ b/src/prompts/fuzzy_select.rs
@@ -81,8 +81,8 @@ impl FuzzySelect<'_> {
     /// Adds multiple items to the fuzzy selector.
     pub fn items<T>(&mut self, items: T) -> &mut Self
     where
-        T::Item: ToString,
         T: IntoIterator,
+        T::Item: ToString,
     {
         for item in items {
             self.items.push(item.to_string());
--- a/src/prompts/fuzzy_select.rs
+++ b/src/prompts/fuzzy_select.rs
@@ -81,8 +81,8 @@ impl FuzzySelect<'_> {
     /// Adds multiple items to the fuzzy selector.
     pub fn items<T>(&mut self, items: T) -> &mut Self
     where
+        T::Item: ToString,
         T: IntoIterator,
-        T::Item: std::string::ToString,
     {
         for item in items {
             self.items.push(item.to_string());
--- a/src/prompts/fuzzy_select.rs
+++ b/src/prompts/fuzzy_select.rs
@@ -79,7 +79,11 @@ impl FuzzySelect<'_> {
     }
 
     /// Adds multiple items to the fuzzy selector.
-    pub fn items<T: ToString>(&mut self, items: &[T]) -> &mut Self {
+    pub fn items<T>(&mut self, items: T) -> &mut Self
+    where
+        T: IntoIterator,
+        T::Item: std::string::ToString,
+    {
         for item in items {
             self.items.push(item.to_string());
         }
--- a/src/prompts/multi_select.rs
+++ b/src/prompts/multi_select.rs
@@ -93,8 +93,8 @@ impl MultiSelect<'_> {
     /// Adds multiple items to the selector.
     pub fn items<T>(&mut self, items: T) -> &mut Self
     where
-        T::Item: ToString,
         T: IntoIterator,
+        T::Item: ToString,
     {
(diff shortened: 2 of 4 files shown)


src/prompts/fuzzy_select.rs
```diff
@@ -79,7 +79,11 @@ impl FuzzySelect<'_> {
     }
 
     /// Adds multiple items to the fuzzy selector.
-    pub fn items<T: ToString>(&mut self, items: &[T]) -> &mut Self {
+    pub fn items<T>(&mut self, items: T) -> &mut Self
+    where
+        T: IntoIterator,
+        T::Item: std::string::ToString,
```
pksunkara, line 85: Do we need `std::string`? Similarily below

src/prompts/multi_select.rs
```diff
@@ -100,8 +104,12 @@ impl MultiSelect<'_> {
     }
 
     /// Adds multiple items to the selector with checked state
-    pub fn items_checked<T: ToString>(&mut self, items: &[(T, bool)]) -> &mut Self {
-        for &(ref item, checked) in items {
+    pub fn items_checked<T, I>(&mut self, items: T) -> &mut Self
+    where
+        I: std::string::ToString,
+        T: IntoIterator<Item = (I, bool)>,
```
pksunkara, line 110: Reorder these 2 lines please.

waterlubber:
[54181b0](https://github.com/mitsuhiko/dialoguer/pull/223/commits/54181b0c87755d3b2d32f75c8d18215fa6db1b98) should fix these comments. Thanks!

waterlubber:
[54181b0](https://github.com/mitsuhiko/dialoguer/pull/223/commits/54181b0c87755d3b2d32f75c8d18215fa6db1b98) should fix these problems. By mistake, I reordered all the other terms; so [2e280e6](https://github.com/mitsuhiko/dialoguer/pull/223/commits/2e280e6db877142e71895e7b97709f0de3030bee) fixes that. Sorry about that.

pksunkara:
> Existing code should continue to work, as &[T] implements IntoIterator<Item: T> by default.

Can you try it locally? I remember we had issues about that in Clap.

```rs
// Try giving it directly like this
.items(&["one", "two"])
```

https://github.com/clap-rs/clap/pull/2857
https://github.com/clap-rs/clap/issues/2870
https://github.com/clap-rs/clap/pull/4081

I think this is a breaking change. (I don't mind it but I wanted to confirm)

waterlubber:
I just ran some basic tests to see if things still worked in an intuitive manner:
```rust
use crate::{MultiSelect, Select, Sort};

    #[cfg(feature = "fuzzy-select")]
    #[test]
    fn fuzzy_test() {
        use crate::FuzzySelect;
        let s1 = FuzzySelect::new()
            .items(&["ham", "muenster", "bread"])
            .with_prompt("Select Something?")
            .interact();
        let items = vec![("foo", true), ("bar", false), ("baz", true)];
        let s2 = MultiSelect::new().items_checked(&items).interact();
        let s22 = MultiSelect::new()
            .items_checked(&[("test", true), ("test2", false)])
            .interact();
        let s3 = Select::new().items(["a"]).interact();
        let items_2 = items.iter().map(|(a, b)| a);
        let s4 = Sort::new().items(items_2).interact();
    }
    ``` 
    
They do, but there is a small problem with MultiSelect. IntoIterator on Vec gives references to tuples (`&(T, bool)`) instead of tuples (`(T, bool)`). 

I can slightly change the function definition to fix this:
```rust
 pub fn items_checked<'a, T, I>(&mut self, items: T) -> &mut Self
    where
        T: IntoIterator<Item = &'a (I, bool)>,
        I: ToString + 'a,
    {
        for (item, checked) in items {
            self.items.push(item.to_string());
            self.defaults.push(*checked);
        }
        self
    }
    ```
    
Which makes it work very well with the tests above. _This_ might be a breaking change, but I'm not entirely sure. Rust might automatically coerce these types. If this is an acceptable solution, I can go ahead and commit this change to the PR.

PR #224, opened 2022-10-24
Adding post completion text for input

robinbernon:
Adding option to be able to change the prompt text after an input has been given.

### DIFF
--- a/src/prompts/input.rs
+++ b/src/prompts/input.rs
@@ -39,6 +39,7 @@ use console::{Key, Term};
 /// ```
 pub struct Input<'a, T> {
     prompt: String,
+    post_completion_text: Option<String>,
     report: bool,
     default: Option<T>,
     show_default: bool,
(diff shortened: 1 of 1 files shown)


pksunkara (commented):
Like the idea but I think the name needs to be better. Any other ideas?

robinbernon:
> Like the idea but I think the name needs to be better. Any other ideas?

Only thing that could be more clear here is something like - "with_post_completion_prompt_change". Would you be happy to merge the PR if I change the method to this instead?

pksunkara (commented):
Hmm.. would your use case be solved by turning off report and then using `TermThemeRenderer` yourself?

robinbernon:
> Hmm.. would your use case be solved by turning off report and then using `TermThemeRenderer` yourself?

How would that be possible? Also if it maybe was the case (which I can't see), wouldn't it be best to offer this as a first class feature as it is pretty useful?

pksunkara (approved):

PR #229, opened 2022-11-09
Return Err on Confirm's read_key when not a terminal

### LINKED ISSUE #148: `Confirm::_interact_on` infinite loop when not a tty
[`Confirm::_interact_on` calls `Term::read_key` in a loop](https://github.com/mitsuhiko/dialoguer/blob/f9c067bfe711d5a52cb6bac0c128d046d3ade594/src/prompts/confirm.rs#L185) but doesn't break out of the loop if the return value is `Key::Unknown`, which `read_key` returns if not connected to a terminal. This causes an infinite loop.

DanteMarshal:
Resolves #148

### DIFF
--- a/src/prompts/confirm.rs
+++ b/src/prompts/confirm.rs
@@ -206,6 +206,12 @@ impl Confirm<'_> {
                     Key::Escape | Key::Char('q') if allow_quit => {
                         value = None;
                     }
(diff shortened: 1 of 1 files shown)


DanteMarshal:
What are these issues ?

PR #233, opened 2022-12-09
Fix crash in fuzzy select

cd-work:
Currently when using the fuzzy select without any defaults, the default would be set to `!0`. This is problematic since this default value is used upon pressing enter to index into the provided options and cause an out of bounds.

This patch fixes this issue by using a proper `Option` rather than encoding a missing default into an integer, allowing for distinct treatment of the scenario where the selection has not been set yet.

### DIFF
--- a/src/prompts/fuzzy_select.rs
+++ b/src/prompts/fuzzy_select.rs
@@ -34,7 +34,7 @@ use std::{io, ops::Rem};
 /// ```
 
 pub struct FuzzySelect<'a> {
-    default: usize,
+    default: Option<usize>,
     items: Vec<String>,
     prompt: String,
     report: bool,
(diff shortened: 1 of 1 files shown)


pksunkara (approved):

cd-work:
Seems like it's been a while since the last release. Is there a plan for a 0.10.3 or are you shooting for a 0.11.0 with the remaining tasks documented in the corresponding milestones?

cd-work:
I hate to nag like this, but this PR has been merged for 3 weeks and the last release has been 5 months ago. I'd really appreciate it if some of these fixes could be made available to crates.io users.

If there's any specific issues blocking a version release to crates.io, please let me know and I can look into helping out with them.

pksunkara:
The tasks in milestone are irrelevant. If we wanna release now, we need 0.11.0 because of breaking backward compat.

PR #234, opened 2022-12-11
Flush terminal when initial text is used

data-retriever:
This is similar to https://github.com/mitsuhiko/dialoguer/pull/232

The issue only happens on buffered terminals


pksunkara (commented):
Please fix the indentation.

pksunkara (approved):

Issue #170, opened 2022-02-03
Unable to pipe `yes` into a confirmation


jgrund:
Hi, 

When wanting to bypass a confirmation in an automated way, it does not appear possible to use the `yes` command, I.E:

```sh
yes | some_dialoguer_cmd
```

Will just hang

ip1981:
Probably it could have a code path for the case when stdin is not a tty.

sycured:
Have you found a solution?

I've the same issue and tomorrow, I'll try to use expect

LeoDog896:
#82 using rustyline may be able to solve this issue.

PR #235, opened 2022-12-23
rustyline migration

LeoDog896:
This PR integrates rusty line into dialoguer, bringing in a number of features that aim to solve #170 and various UTF-8 issues.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ completion = []
 
 [dependencies]
 console = "0.15.0"
+rustyline = "10.0.0"
 tempfile = { version = "3", optional = true }
 zeroize = { version = "1.1.1", optional = true }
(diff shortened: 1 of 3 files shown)


pksunkara:
But aren't you removing quite a few interaction related features?

LeoDog896:
Thus why I left this as a draft (though I should have specified this earlier) - I am currently reimplementing history and completion.

pksunkara:
FWIW, https://github.com/mitsuhiko/dialoguer/pull/213 is an alternate attempt to fix this. Unfortunately, all of these issues are blocked on having a proper test suite first.

LeoDog896:
Thanks for letting me know! On top of that, it also seems that rustyline doesn't have a [`History` trait implemented yet](https://github.com/kkawakam/rustyline/pull/604) -- until that PR is finished, the only alternative workaround is hacky, so it's better to wait.

PR #238, opened 2022-12-29
Fix: Handle executable with spaces in name

arxanas:
This was reported by a user of my project at https://github.com/arxanas/git-branchless/issues/718. In their case, they are running Windows, and the path to their executable naturally includes the substring `Program Files`, which has spaces in it. These spaces are not handled properly by `dialoguer`.

An alternative would be to offer an option to disable splitting on space, which would then allow the caller to manually parse the executable string on their own, in the event that it contains spaces; or we could add an alternative method to `Editor` (and update the internal state) so that we can construct it with a string that might contain both the executable and arguments, which will then do the space-splitting behavior.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ console = "0.15.0"
 tempfile = { version = "3", optional = true }
 zeroize = { version = "1.1.1", optional = true }
 fuzzy-matcher = { version = "0.3.7", optional = true }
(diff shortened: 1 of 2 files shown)


pksunkara (approved):

arxanas:
@pksunkara Please let me know when a new release or prerelease version is published, so that I can update the dependency for my package.

Issue #241, opened 2023-01-16
Feature: Multi Fuzzy Select


RobWalt:
In the [`glab`](https://docs.gitlab.com/ee/integration/glab/) CLI tool, there are dialogues where you can select labels for issues for example. These act a lot like the `MultiSelect` prompt of this crate here, but they have the bonus of including a fuzzy search.

You can basically:
- type a work to refine the search
- toggle selection with space bar
- confirm the selction with `Enter`

I would love to see something similar in this lib here.

mrchilliballs:
May I ask why you closed you pull request?

RobWalt:
@mrchilliballs I didn't feel like I wanted to finish that. It's still there so feel free to pick the PR up. The changes aren't too technical

PR #242, opened 2023-01-16
Feat: multi fuzz selection

RobWalt:
Reopening a refined version of the multi fuzz selection tool I tried to implement.

Solves #241 

Here is a small demo:

dialoguer

### DIFF
--- a/examples/multi_fuzzy_select.rs
+++ b/examples/multi_fuzzy_select.rs
@@ -37,6 +37,7 @@ fn main() {
         .with_prompt("Pick your food")
         .items(&multiselected[..])
         .defaults(&defaults[..])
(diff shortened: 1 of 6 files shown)


RobWalt:
Just did a little cleanup, improving the code that I just copy-pasted from the fuzzy-select prompt.

Also made a slight adjustment to the behavior of the cursor when a selection is checked. Now the cursor stays at it's position and isn't reset to position 0.

dialoguer

MrLarssonJr:
This looks neat, I was looking for the same capability myself.

MrLarssonJr:
Ah, thanks for the recommendation! From a first glance it looks great!

PR #245, opened 2023-02-22
Fix panic in fuzzy-select when using non-ASCII characters

### LINKED ISSUE #181: Fuzzy Select: Special characters lead to panic

stormshield-kg:
Fixes #181.

Use the [`str::char_indices()`](https://doc.rust-lang.org/std/primitive.str.html#method.char_indices) to keep valid byte offsets in memory.

Also add handling of  the `Delete` key.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@
 * `Input` values that are invalid are now also stored in `History`
 * Resolve some issues with cursor positioning in `Input` when using `utf-8` characters
 * Correct page is shown when default selected option is not on the first page for `Select`
+* Fix panic in `FuzzySelect` when using non-ASCII characters
+* Add handling of `Delete` key for `FuzzySelect`
 
 ### Breaking
 
@@ -26,6 +28,7 @@
 * Prompt interaction functions now take `self` instead of `&self`
 * Prompt interaction functions and other operations now return `dialouger::Result` instead of `std::io::Result`
 * Rename `Validator` to `InputValidator`
+* The trait method `Theme::format_fuzzy_select_prompt()` now takes a byte position instead of a cursor position in order to support UTF-8.
 
 ## 0.10.4
 
--- a/src/prompts/fuzzy_select.rs
+++ b/src/prompts/fuzzy_select.rs
@@ -343,6 +343,10 @@ impl FuzzySelect<'_> {
                     search_term.remove(byte_indices[cursor]);
                     term.flush()?;
                 }
+                (Key::Del, _, _) if cursor < byte_indices.len() - 1 => {
+                    search_term.remove(byte_indices[cursor]);
+                    term.flush()?;
+                }
                 (Key::Char(chr), _, _) if !chr.is_ascii_control() => {
                     search_term.insert(byte_indices[cursor], chr);
                     cursor += 1;
--- a/src/prompts/fuzzy_select.rs
+++ b/src/prompts/fuzzy_select.rs
@@ -195,7 +195,7 @@ impl FuzzySelect<'_> {
 
     fn _interact_on(self, term: &Term, allow_quit: bool) -> Result<Option<usize>> {
         // Place cursor at the end of the search term
-        let mut position = self.initial_text.len();
+        let mut cursor = self.initial_text.chars().count();
         let mut search_term = self.initial_text.to_owned();
 
         let mut render = TermThemeRenderer::new(term, self.theme);
@@ -224,8 +224,15 @@ impl FuzzySelect<'_> {
         let mut vim_mode = false;
 
         loop {
+            let mut byte_indices = search_term
+                .char_indices()
+                .map(|(index, _)| index)
+                .collect::<Vec<_>>();
+
+            byte_indices.push(search_term.len());
+
             render.clear()?;
-            render.fuzzy_select_prompt(self.prompt.as_str(), &search_term, position)?;
+            render.fuzzy_select_prompt(self.prompt.as_str(), &search_term, byte_indices[cursor])?;
 
             // Maps all items to a tuple of item and its match score.
             let mut filtered_list = self
@@ -304,14 +311,14 @@ impl FuzzySelect<'_> {
                     }
                     term.flush()?;
                 }
-                (Key::ArrowLeft, _, _) | (Key::Char('h'), _, true) if position > 0 => {
-                    position -= 1;
+                (Key::ArrowLeft, _, _) | (Key::Char('h'), _, true) if cursor > 0 => {
+                    cursor -= 1;
                     term.flush()?;
                 }
                 (Key::ArrowRight, _, _) | (Key::Char('l'), _, true)
-                    if position < search_term.len() =>
+                    if cursor < byte_indices.len() - 1 =>
                 {
-                    position += 1;
+                    cursor += 1;
                     term.flush()?;
                 }
                 (Key::Enter, Some(sel), _) if !filtered_list.is_empty() => {
@@ -331,14 +338,14 @@ impl FuzzySelect<'_> {
                     term.show_cursor()?;
                     return Ok(sel_string_pos_in_items);
                 }
-                (Key::Backspace, _, _) if position > 0 => {
-                    position -= 1;
-                    search_term.remove(position);
+                (Key::Backspace, _, _) if cursor > 0 => {
+                    cursor -= 1;
+                    search_term.remove(byte_indices[cursor]);
                     term.flush()?;
                 }
                 (Key::Char(chr), _, _) if !chr.is_ascii_control() => {
-                    search_term.insert(position, chr);
-                    position += 1;
+                    search_term.insert(byte_indices[cursor], chr);
+                    cursor += 1;
                     term.flush()?;
                     sel = Some(0);
                     starting_row = 0;
--- a/src/theme/colorful.rs
+++ b/src/theme/colorful.rs
@@ -408,31 +408,24 @@ impl Theme for ColorfulTheme {
         f: &mut dyn fmt::Write,
         prompt: &str,
         search_term: &str,
-        cursor_pos: usize,
+        bytes_pos: usize,
     ) -> fmt::Result {
         if !prompt.is_empty() {
             write!(
                 f,
                 "{} {} ",
-                &self.prompt_prefix,
+                self.prompt_prefix,
                 self.prompt_style.apply_to(prompt)
             )?;
         }
 
-        if cursor_pos < search_term.len() {
-            let st_head = search_term[0..cursor_pos].to_string();
-            let st_tail = search_term[cursor_pos + 1..search_term.len()].to_string();
-            let st_cursor = self
-                .fuzzy_cursor_style
-                .apply_to(search_term.to_string().chars().nth(cursor_pos).unwrap());
-            write!(
-                f,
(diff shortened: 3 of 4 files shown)


Gordon01:
I tested this PR and it indeed fixes the problem with non-ASCII characters. 

However, I believe that `format_fuzzy_select_prompt()` may be the only one that needs to be fixed to correctly support UTF-8.
Specifically replace indexing with `.chars()` method as described here: https://stackoverflow.com/questions/48642342/how-to-get-the-last-character-of-a-str

src/prompts/fuzzy_select.rs
```diff
@@ -165,9 +165,18 @@ impl FuzzySelect<'_> {
...
         let mut search_term = self.initial_text.to_owned();
 
+        let mut position_indices = search_term
+            .char_indices()
+            .map(|(index, _)| index)
+            .collect::<Vec<_>>();
+
+        position_indices.push(search_term.len());
+
+        let mut char_index = position_indices.len() - 1;
```
Gordon01, line 178: A correct iteration over chars should be used in rendering instead.
As described here: https://stackoverflow.com/questions/24542115/how-to-index-a-string-in-rust

Gordon01 (changes requested):
Thanks for this PR

src/prompts/fuzzy_select.rs
stormshield-kg: With the `.chars()` iterator, you cannot go the left of a string, only to the right. I used a `Vec` of char indices in order to be able to remove a letter in the middle of the text, if the cursor is not at the end.

Gordon01:
I believe we don't need an optimization for `search_term` indexes. 
It's small (will fin into the L1 cache) so recalculating it in a loop will not affect performance, but it makes the code simpler.

<details><summary>Details</summary>
<p>

```diff
diff --git a/src/prompts/fuzzy_select.rs b/src/prompts/fuzzy_select.rs
index 39824be..3f83947 100644
--- a/src/prompts/fuzzy_select.rs
+++ b/src/prompts/fuzzy_select.rs
@@ -195,18 +195,9 @@ impl FuzzySelect<'_> {
 
     fn _interact_on(self, term: &Term, allow_quit: bool) -> Result<Option<usize>> {
         // Place cursor at the end of the search term
-        let mut bytes_position = self.initial_text.len();
+        let mut cursor = self.initial_text.chars().count();
         let mut search_term = self.initial_text.to_owned();
 
-        let mut position_indices = search_term
-            .char_indices()
-            .map(|(index, _)| index)
-            .collect::<Vec<_>>();
-
-        position_indices.push(search_term.len());
-
-        let mut char_index = position_indices.len() - 1;
-
         let mut render = TermThemeRenderer::new(term, self.theme);
         let mut sel = self.default;
 
@@ -233,8 +224,14 @@ impl FuzzySelect<'_> {
         let mut vim_mode = false;
 
         loop {
+            let mut byte_indices = search_term
+                .char_indices()
+                .map(|(index, _)| index)
+                .collect::<Vec<_>>();
+            byte_indices.push(search_term.len());
+
             render.clear()?;
-            render.fuzzy_select_prompt(self.prompt.as_str(), &search_term, bytes_position)?;
+            render.fuzzy_select_prompt(self.prompt.as_str(), &search_term, byte_indices[cursor])?;
 
             // Maps all items to a tuple of item and its match score.
             let mut filtered_list = self
@@ -313,16 +310,14 @@ impl FuzzySelect<'_> {
                     }
                     term.flush()?;
                 }
-                (Key::ArrowLeft, _, _) | (Key::Char('h'), _, true) if char_index > 0 => {
-                    char_index -= 1;
-                    bytes_position = position_indices[char_index];
+                (Key::ArrowLeft, _, _) | (Key::Char('h'), _, true) if cursor > 0 => {
+                    cursor -= 1;
                     term.flush()?;
                 }
                 (Key::ArrowRight, _, _) | (Key::Char('l'), _, true)
-                    if char_index < position_indices.len() - 1 =>
+                    if cursor < byte_indices.len() - 1 =>
                 {
-                    char_index += 1;
-                    bytes_position = position_indices[char_index];
+                    cursor += 1;
                     term.flush()?;
                 }
                 (Key::Enter, Some(sel), _) if !filtered_list.is_empty() => {
@@ -342,44 +337,18 @@ impl FuzzySelect<'_> {
                     term.show_cursor()?;
                     return Ok(sel_string_pos_in_items);
                 }
-                (Key::Backspace, _, _) if char_index > 0 => {
-                    let old_position = bytes_position;
-
-                    char_index -= 1;
-                    bytes_position = position_indices[char_index];
-
-                    let char_len = old_position - bytes_position;
-                    position_indices.remove(char_index);
-                    for index in &mut position_indices[char_index..] {
-                        *index -= char_len;
-                    }
-
-                    search_term.replace_range(bytes_position..old_position, "");
+                (Key::Backspace, _, _) if cursor > 0 => {
+                    cursor -= 1;
+                    search_term.remove(byte_indices[cursor]);
                     term.flush()?;
                 }
-                (Key::Del, _, _) if char_index < position_indices.len() - 1 => {
-                    let char_len = position_indices[char_index + 1] - bytes_position;
-                    position_indices.remove(char_index + 1);
-
-                    for index in &mut position_indices[char_index + 1..] {
-                        *index -= char_len;
-                    }
-
-                    search_term.replace_range(bytes_position..bytes_position + char_len, "");
+                (Key::Del, _, _) if cursor < byte_indices.len() - 1 => {
+                    search_term.remove(byte_indices[cursor]);
                     term.flush()?;
                 }
                 (Key::Char(chr), _, _) if !chr.is_ascii_control() => {
-                    let char_len = chr.len_utf8();
-                    let old_position = bytes_position;
-                    bytes_position += char_len;
-                    position_indices.insert(char_index, old_position);
-
-                    char_index += 1;
-                    for index in &mut position_indices[char_index..] {
-                        *index += char_len;
-                    }
-
-                    search_term.insert(old_position, chr);
+                    search_term.insert(byte_indices[cursor], chr);
+                    cursor += 1;
                     term.flush()?;
                     sel = Some(0);
                     starting_row = 0;

```  

</p>
</details>

src/prompts/fuzzy_select.rs
```diff
@@ -331,14 +342,44 @@ impl FuzzySelect<'_> {
...
+                }
+                (Key::Del, _, _) if char_index < position_indices.len() - 1 => {
+                    let char_len = position_indices[char_index + 1] - bytes_position;
+                    position_indices.remove(char_index + 1);
+
+                    for index in &mut position_indices[char_index + 1..] {
+                        *index -= char_len;
+                    }
+
+                    search_term.replace_range(bytes_position..bytes_position + char_len, "");
```
Gordon01, line 368: We should put indexes recalculation in a loop and ditch this code.
I believe that it's an unnecessary optimisation since `search_term` is tiny but it complicates those functions a lot.

src/theme/colorful.rs
```diff
@@ -408,31 +408,21 @@ impl Theme for ColorfulTheme {
...
         if !prompt.is_empty() {
-            write!(
-                f,
-                "{} {} ",
-                &self.prompt_prefix,
-                self.prompt_style.apply_to(prompt)
-            )?;
+            let prompt_prefix = &self.prompt_prefix;
+            let styled_prompt = self.prompt_style.apply_to(prompt);
+            write!(f, "{prompt_prefix} {styled_prompt} ")?;
```
Gordon01, line 416: Those bindings aren't used anywhere else. Let's leave this code intact.

Gordon01 (changes requested):
Please consider the attached diff.

Also, please update the changelog.

Gordon01:
Just realized that `Input` builds a `Vec` of chars. That's also a way for code simplification. https://github.com/console-rs/dialoguer/blob/master/src/prompts/input.rs#L321

Gordon01:
Let's add Delete key handling to Input too, since it also lacks this functionality.

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@
 * `Input` values that are invalid are now also stored in `History`
 * Resolve some issues with cursor positioning in `Input` when using `utf-8` characters
 * Correct page is shown when default selected option is not on the first page for `Select`
+* Fix panic in fuzzy-select when using non-ASCII characters
+* Add handling of Delete key for fuzzy-select
```
Gordon01, line 20: Use `FuzzySelect` instead of fuzzy-select

Gordon01 (approved):
Looks great, thank you!

Gordon01 (commented):
LGTM

src/prompts/fuzzy_select.rs
```diff
@@ -304,14 +311,14 @@ impl FuzzySelect<'_> {
...
                 }
-                (Key::ArrowLeft, _, _) | (Key::Char('h'), _, true) if position > 0 => {
-                    position -= 1;
+                (Key::ArrowLeft, _, _) | (Key::Char('h'), _, true) if cursor > 0 => {
+                    cursor -= 1;
                     term.flush()?;
                 }
                 (Key::ArrowRight, _, _) | (Key::Char('l'), _, true)
-                    if position < search_term.len() =>
+                    if cursor < byte_indices.len() - 1 =>
```
pksunkara, line 319: I haven't tried it locally but is the `-1` needed?

src/prompts/fuzzy_select.rs
```diff
@@ -331,14 +338,18 @@ impl FuzzySelect<'_> {
...
                 }
-                (Key::Backspace, _, _) if position > 0 => {
-                    position -= 1;
-                    search_term.remove(position);
+                (Key::Backspace, _, _) if cursor > 0 => {
+                    cursor -= 1;
+                    search_term.remove(byte_indices[cursor]);
+                    term.flush()?;
+                }
+                (Key::Del, _, _) if cursor < byte_indices.len() - 1 => {
```
pksunkara, line 346: Same question about `-1`.

src/theme/mod.rs
```diff
@@ -253,20 +253,13 @@ pub trait Theme {
         f: &mut dyn fmt::Write,
         prompt: &str,
         search_term: &str,
-        cursor_pos: usize,
+        bytes_pos: usize,
```
pksunkara, line 256: This is a breaking change that needs to be talked about in the changelog.

src/prompts/fuzzy_select.rs
Gordon01: Yes, because byte_indices contains one additional element: `byte_indices.push(search_term.len());`

src/prompts/fuzzy_select.rs
Gordon01: Same here

src/theme/mod.rs
Gordon01: Are you sure? This is a parameter of the private method.

stormshield-kg: It is breaking since this is a public trait method. I have updated the changelog.

pksunkara (approved):

PR #252, opened 2023-04-15
Drop .clippy.toml

LingMan:
Clippy can read the `rust-version` field from the Cargo.toml nowadays. No need to duplicate this information.


pksunkara:
Which  clippy version is that? We install clippy on rust 1.59.0 in CI

LingMan:
The MSRV is 1.59, so if you use 1.59 for linting, this setting has no effect anyway. Not like it can filter out lints it doesn't have yet.

To answer your question: It reads `rust-version` since Rust 1.64 ([changelog entry](https://github.com/rust-lang/rust-clippy/blob/master/CHANGELOG.md#rust-164)).

Although my PR polishing the CI script does contain a commit to use Rust stable instead of 1.59 for linting.

PR #254, opened 2023-04-15
Polish CI

LingMan:
- The big item here is switching away from actions-rs since the entire organization is unmaintained and the actions are starting to fall apart. 
- The second item is switching the lint job from running with the MSRV to stable, so dialoguer doesn't miss out on newer lints.
- The other two commits are just nits without any functional change.
 
See the individual commits for details.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,5 @@
 [alias]
 format       = "fmt"
 format-check = "fmt --check"
-lint         = "clippy --all-targets --all-features --examples --tests -- -D warnings"
+lint         = "clippy --all-targets --all-features -- -D warnings"
 test-cover   = "llvm-cov --all-features --lcov --output-path lcov.info"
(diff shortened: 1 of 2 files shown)


pksunkara:
Can you give me an example on where the actions are falling apart? One advantage of cargo action is that it comments the warnings and errors on GitHub.

LingMan:
Quoting from the relevant commit message:

> The actions still use Node 12 and the `set-output` command, both of which are deprecated and will
be removed by Github Actions by Summer 2023. [1][2]
> [...]
> [1] https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/
> [2] https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

pksunkara (approved):

pksunkara:
I think this needs rebase.

LingMan:
Rebased. Am I assuming correctly that the other two comments you left about the examples and the tag have resolved themselves? Can't find them in the web UI and you approved the PR, so it look like you deleted them.

PR #258, opened 2023-05-13
Added recordings of examples to the readme

FelixSelter:
When programmers search for a great library they often find multiple and need to decide which one is worth being used. A good readme showing the possibilities of a library helps them to quickly identify if it matches their situation. Forcing them to download the library and to run the examples is unnecessarily annoying


pksunkara:
Is linking to screenshots in https://github.com/termapps/enquirer enough?

FelixSelter:
Seems like enquirer has not an example for every possible dialog. Also Its somewhat a different tool. I do not see any reason to not create a few more svgs for this repo. The advantage is you know which example produced them. People might not even get that dialouger is a library if they see this enquirer specific syntax. Is there anything wrong with the recordings? Because the onlyproblem I see are a few more bytes github has to store somewhere.

pksunkara:
I guess, I am trying to figure out what exactly are we trying to provide with these gifs? Examples? Or comprehensive behavior showcase? Or something else?

FelixSelter:
My intention was to show examples of all possible dialogs that this library helps to create. Therefore anyone looking at the readme knows if the library fits his usecase without downloading everything and running the examples himself. You can quickly get an overview about all possibilities and investigate in what you need.

pksunkara:
> all possible dialogs that this library helps to create.

> about all possibilities and investigate in what you need.

Yeah, but since the dialogs can have different behaviors based on different options, are you talking about just one behavior as a gif for each dialog or each behavior for each dialog in the gif?

FelixSelter:
I personally would prefer one gif for every dialogue showing all behaviours of that single dialogue. Maybe they could even be generated automatically?

pksunkara:
> I personally would prefer one gif for every dialogue showing all behaviors of that single dialogue. 

Each dialogue can have a lot of different behaviors.

> Maybe they could even be generated automatically

And yes, they would need to be generated automatically. Otherwise, it's not scalable.

FelixSelter:
I guess this is at least a temporary start

joshka:
I came across this library just now. It's quite neat, and I had the same exact thought that it needs examples to get an idea of what this looks like. I cloned a copy and was wowed enough by the examples to make use of it immediately, but if I hadn't I probably would have never done so.

Regarding putting a large amount of examples on the readme itself, perhaps it would be worth showing a small static example of each of the prompts, and having these more in depth examples on a seperate examples doc.

Regarding formatting, I'd suggest replacing the markdown for the svgs with img tags to be able to specify a width for each example. Also consider finding a way to get rid of the noise at the top of each example.

P.s. what's the tool you're using to create the SVGs? I've used Asciinema for this sort of thing in the past, which I find is nice because it comes with the ability to pause / scrub the demo, but this is still pretty cool for things that don't need that as much (and that renders automatically on gh).

pksunkara:
https://github.com/marionebl/svg-term-cli is what I use normally.

FelixSelter:
I used term2svg. Getting rid of the noise can be done easily. term2svg files can also be edited with ascinema edit if you are using this tool. Maybe I can work on this in a week but I have very limited time at the moment

FelixSelter:
@pksunkara Can you give me an example of one dialogue and what you would expect to be generated.

joshka:
I created an alternative to this PR that gets rid of all the terminal cruft, leaving just the relevant output: https://github.com/console-rs/dialoguer/pull/317. Direct link to view the README: https://github.com/console-rs/dialoguer/blob/a9428d4e6601949112fe2194b9ae64c70d8d6f99/README.md.

PR #259, opened 2023-05-20
Add vim mode to `FuzzySelect`

damymetzke:
This pull request introduces a vim mode to `FuzzySelect`.
Vim mode works as follows:

1. The user can press Escape to enter vim mode
2. The user can use hjkl to navigate
3. The user can still select with Enter, or press i/a to enter more text

This is based on the vim mode from <https://github.com/go-survey/survey>.

To achieve this I've added the following changes:

- Keep track of `vim_mode`.
- Add support for hjkl keys, only when `vim_mode` is true.
- Add the `FuzzySelect::enable_vim_mode` method.

These changes have the following benefits:

- They allow vim users to use vim bindings, even in a fuzzy select.
- The behavior is opt-in, so other users won't even notice.
- The behavior is the same as many other tools, one example being Telescope for Neovim.

I hope my contribution can be of use!

### DIFF
--- a/src/prompts/fuzzy_select.rs
+++ b/src/prompts/fuzzy_select.rs
@@ -258,7 +258,7 @@ impl FuzzySelect<'_> {
                 (Key::Escape, _, false) if self.enable_vim_mode => {
                     vim_mode = true;
                 }
-                (Key::Escape, _, _) if allow_quit => {
+                (Key::Escape, _, false) | (Key::Char('q'), _, true) if allow_quit => {
                     if self.clear {
                         render.clear()?;
                         term.flush()?;
--- a/src/prompts/fuzzy_select.rs
+++ b/src/prompts/fuzzy_select.rs
@@ -38,6 +38,7 @@ pub struct FuzzySelect<'a> {
     report: bool,
     clear: bool,
     highlight_matches: bool,
+    enable_vim_mode: bool,
     max_length: Option<usize>,
     theme: &'a dyn Theme,
     /// Search string that a fuzzy search with start with.
@@ -115,6 +116,17 @@ impl FuzzySelect<'_> {
     /// The default is to highlight the indices
     pub fn highlight_matches(mut self, val: bool) -> Self {
         self.highlight_matches = val;
+        self
+    }
+
+    /// Indicated whether to allow the use of vim mode
+    ///
+    /// Vim mode can be entered by pressing Escape.
+    /// This then allows the user to navigate using hjkl.
+    ///
+    /// The default is to disable vim mode.
+    pub fn enable_vim_mode(mut self, val: bool) -> Self {
+        self.enable_vim_mode = val;
         self
     }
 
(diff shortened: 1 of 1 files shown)


src/prompts/fuzzy_select.rs
```diff
@@ -118,6 +119,17 @@ impl FuzzySelect<'_> {
         self
     }
 
+    /// Indicated whether to allow the use of vim mode
+    ///
+    /// Vim mode can be entered by pressing Escape.
+    /// This then allows the user to navigate using hjkl.
+    ///
+    /// The default is to disable vim mode.
+    pub fn enable_vim_mode(mut self, val: bool) -> Self {
```
pksunkara, line 128:
```suggestion
    pub fn vim_mode(mut self, val: bool) -> Self {
```

pksunkara (commented):
This won't work because you are losing the ability to write some characters in search term.

src/prompts/fuzzy_select.rs
```diff
@@ -240,16 +254,24 @@ impl FuzzySelect<'_> {
...
                         render.clear()?;
                         term.flush()?;
                     }
                     term.show_cursor()?;
                     return Ok(None);
                 }
-                (Key::ArrowUp | Key::BackTab, _) if !filtered_list.is_empty() => {
+                (Key::Escape, _, false) if self.enable_vim_mode => {
+                    vim_mode = true;
+                }
```
pksunkara, line 268: How do I `allow_quit` when `enable_vim_mode`?

src/prompts/fuzzy_select.rs
```diff
@@ -305,17 +331,18 @@ impl FuzzySelect<'_> {
...
                     term.flush()?;
                 }
-                (Key::Char(chr), _) if !chr.is_ascii_control() => {
+                (Key::Char(chr), _, _) if !chr.is_ascii_control() => {
                     search_term.insert(position, chr);
                     position += 1;
                     term.flush()?;
                     sel = Some(0);
                     starting_row = 0;
+                    vim_mode = false;
```
pksunkara, line 345: Do you need this line here?

pksunkara (changes requested):

src/prompts/fuzzy_select.rs
damymetzke: Good point, I've implemented it so that you can quit by pressing Escape again in vim mode.

src/prompts/fuzzy_select.rs
damymetzke: I guess not, I removed it.

src/prompts/fuzzy_select.rs
```diff
@@ -209,6 +221,8 @@ impl FuzzySelect<'_> {
 
         term.hide_cursor()?;
 
+        let mut vim_mode = false;
```
pksunkara, line 224: Shouldn't this default to `enable_vim_mode`?

src/prompts/fuzzy_select.rs
```diff
@@ -240,16 +254,24 @@ impl FuzzySelect<'_> {
             }
             term.flush()?;
 
-            match (term.read_key()?, sel) {
-                (Key::Escape, _) if allow_quit => {
+            match (term.read_key()?, sel, vim_mode) {
+                (Key::Escape, _, false) if self.enable_vim_mode => {
+                    vim_mode = true;
+                }
+                (Key::Escape, _, _) if allow_quit => {
```
pksunkara, line 261: Shouldn't this be `q` if in vim_mode?

src/prompts/fuzzy_select.rs
damymetzke: Probably not. enable_vim_mode is meant to fully disable vim mode; the vim_mode variable keeps track of when you are entering text or when vim bindings are used. Typically you'd fuzzy find first, then use vim mode, So I think it should always start with no vim mode.

src/prompts/fuzzy_select.rs
damymetzke: I've seen both used in many programs, so I don't have a preference. I already changed it.

pksunkara (approved):

Issue #268, opened 2023-07-13
BUG - Input with utf8 chars 


finistseb:
When we enter utf8char with input.interact_text(), the behavior is buggy, the cursor moves in not accurate do to a bad string length calculation. 
The tail length is realized with tail.len() (bytes count where 1 utf8 char = 2 bytes) instead of tail.chars().count()

I have proposed a  [PR](https://github.com/console-rs/dialoguer/pull/269)) in order to fix this issue.
I'm a Rust developer for a relatively short time, so me bug analysis may be wrong or incomplete ;-) 

Regards, 
Sebastien

finistseb:
How to reproduce with this simple test program, 
enter the input "aaaaaa" and after insert a "é" in the middle of the string. 
The output is correct, but the display is wrong. 

```
use dialoguer::{theme::ColorfulTheme, Input};

fn main() {
    let color_binding = ColorfulTheme::default();
    let value: String = Input::with_theme(&color_binding)
        .with_prompt("Enter parameter")
        .report(false)
        .allow_empty(false)
        .interact_text().unwrap();

    println!("{}", value);
}
```

PR #269, opened 2023-07-13
[#268] Fix Input tail length calculation (wrong cursor positioning)

finistseb:
Cursor is not correctly positioned where the input string contains utf8 chars.


pksunkara (approved):

francisdb:
Could we get a release containing his patch?

pksunkara:
Unfortunately, it might be hard because I was in the middle of breaking changes. Might have to port it to a different branch.

PR #275, opened 2023-08-30
Change chars for `MultiSelect` items

Gordon01:
Windows Terminal, a widely used terminal emulator for some reason does not support coloring for '✔' char.
'✔️' and '❌' may be used instead. They work down to the previous Windows terminal emulator: `conhost.exe`.

### Windows Terminal 
image
As you can see, selected and unselected items are indistinguishable.

### `conhost.exe`

image
Surprisingly, the old windows emulator is better than new one. 

### VSCode
image

### Windows Terminal 

image

### `conhost.exe`

image

### VSCode

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * Added `dialouger::Result` and `dialouger::Error`
 * Resolve some issues on Windows where pressing shift keys sometimes aborted dialogs.
+* Resolve `MultiSelect` checked and unchecked variants looking the same on Windows. More compatible '✔️' and '❌' used instead of '✔' colors.
 
 ### Breaking
 
--- a/src/theme/colorful.rs
+++ b/src/theme/colorful.rs
@@ -71,8 +71,8 @@ impl Default for ColorfulTheme {
             inactive_item_style: Style::new().for_stderr(),
             active_item_prefix: style("❯".to_string()).for_stderr().green(),
             inactive_item_prefix: style(" ".to_string()).for_stderr(),
-            checked_item_prefix: style("✔".to_string()).for_stderr().green(),
-            unchecked_item_prefix: style("✔".to_string()).for_stderr().black(),
+            checked_item_prefix: style("✔️".to_string()).for_stderr().blue(),
+            unchecked_item_prefix: style("❌".to_string()).for_stderr().red(),
             picked_item_prefix: style("❯".to_string()).for_stderr().green(),
             unpicked_item_prefix: style(" ".to_string()).for_stderr(),
             #[cfg(feature = "fuzzy-select")]


Gordon01:
This fixes https://github.com/console-rs/dialoguer/issues/202 with a well-supported character down to early windows-style `utf-16ish`.

The theme looks good and is consistent on different emulators.

Additional testing is needed for: https://github.com/console-rs/dialoguer/issues/149

pksunkara:
IMHO, it doesn't look good.

Gordon01:
Which new rendering do you think is the worst?

The new VSCode's terminal rendering is questionable.

pksunkara:
The same rendering is also used for Mac and Linux, where the X emoji does not look good.

Gordon01:
What about '☑' '☒' or '✔' '✘' '☓'

src/theme/colorful.rs
```diff
@@ -71,8 +71,8 @@ impl Default for ColorfulTheme {
             inactive_item_style: Style::new().for_stderr(),
             active_item_prefix: style("❯".to_string()).for_stderr().green(),
             inactive_item_prefix: style(" ".to_string()).for_stderr(),
-            checked_item_prefix: style("✔".to_string()).for_stderr().green(),
-            unchecked_item_prefix: style("✔".to_string()).for_stderr().black(),
```
pksunkara, line 75: Try `◌` but don't change the colors.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 * Added `dialouger::Result` and `dialouger::Error`
 * Resolve some issues on Windows where pressing shift keys sometimes aborted dialogs.
+* Resolve `MultiSelect` checked and unchecked variants looking the same on Windows. More compatible '✔️' and '❌' used instead of '✔' colors.
```
pksunkara, line 9: Let's remove the second sentence.

Gordon01:
Linting fails because of: https://github.com/console-rs/dialoguer/pull/209

pksunkara:
Feel free to fix it in this PR

CHANGELOG.md
Gordon01: Done

src/theme/colorful.rs
pksunkara: I checked locally and this looks too small on many fonts. I have tried many other characters; most are too small or too wide. `⬚` looks the best. So, let's use that, and set the color to `magenta`

Gordon01: Ok, I've checked locally too and it's good. Except conhost doesn't know this symbol but renders color correctly.

Gordon01:
image | image
-- | --

image

PR #276, opened 2023-09-05
Count characters in `password` example instead of bytes

Gordon01:
A `password` example contains a typical Rust bug when String's number of bytes is checked instead of characters.

`Password::interact_on()` doc is updated here too.


pksunkara (changes requested):
Please change the examples in docs for password validation too.

Gordon01:
Done

Issue #160, opened 2022-01-07
interact_opt and interact_text_opt for Input


davidanthoff:
Would be great if there was a version of the text input where a user can abort by pressing `Esc`.

DanConwayDev:
I agree. This would be a nice feature.

PR #278, opened 2023-09-10
Input opt

tbergerd:
Hello,
This adds 'interact_text_opt' and 'interact_text_on_opt' for 'Input' prompts (see #160)

Note I did not do the same for 'interact' methods, I don't see how this would work with escape sequences being accepted.

P.S: Is there a specific process to contribute or is opening a pull request fine ?

### DIFF
--- a/src/prompts/input.rs
+++ b/src/prompts/input.rs
@@ -289,6 +289,7 @@ where
     ///
     /// The dialog is rendered on stderr.
     /// This unlike [`interact_text_opt`](Self::interact_text_opt) does not allow to quit with 'Esc'
+    #[inline]
     pub fn interact_text(self) -> Result<T> {
         self.interact_text_on(&Term::stderr())
     }
--- a/src/prompts/input.rs
+++ b/src/prompts/input.rs
@@ -289,12 +289,38 @@ where
     /// while [`interact`](Self::interact) allows virtually any character to be used e.g arrow keys.
     ///
     /// The dialog is rendered on stderr.
+    /// This unlike [`interact_text_opt`](Self::interact_text_opt) does not allow to quit with 'Esc'
     pub fn interact_text(self) -> Result<T> {
         self.interact_text_on(&Term::stderr())
     }
 
+    /// Enables the user to enter a printable ascii sequence and returns the result.
+    ///
+    /// Its difference from [`interact_opt`](Self::interact_opt) is that it only allows ascii characters for string,
+    /// while [`interact_opt`](Self::interact_opt) allows virtually any character to be used e.g arrow keys.
+    ///
+    /// The dialog is rendered on stderr.
+    /// Result contains `Some(T)` if user hit 'Enter' or `None` if user cancelled with 'Esc'
+    #[inline]
+    pub fn interact_text_opt(self) -> Result<Option<T>> {
+        self.interact_text_on_opt(&Term::stderr())
+    }
+
     /// Like [`interact_text`](Self::interact_text) but allows a specific terminal to be set.
-    pub fn interact_text_on(mut self, term: &Term) -> Result<T> {
+    #[inline]
+    pub fn interact_text_on(self, term: &Term) -> Result<T> {
+        Ok(self
+            ._interact_text_on(term, false)?
+            .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "Quit not allowed in this case"))?)
+    }
+
+    /// Like [`interact_text_opt`](Self::interact_text_opt) but allows a specific terminal to be set.
+    #[inline]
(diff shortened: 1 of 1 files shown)


src/prompts/input.rs
```diff
@@ -289,12 +289,38 @@ where
...
     ///
     /// The dialog is rendered on stderr.
+    /// This unlike [`interact_text_opt`](Self::interact_text_opt) does not allow to quit with 'Esc'
     pub fn interact_text(self) -> Result<T> {
         self.interact_text_on(&Term::stderr())
     }
 
+    /// Enables the user to enter a printable ascii sequence and returns the result.
+    ///
+    /// Its difference from [`interact_opt`](Self::interact_opt) is that it only allows ascii characters for string,
```
tbergerd, line 299: oops that's wrong => no interact_opt here

src/prompts/input.rs
```diff
@@ -339,11 +338,6 @@ where
                 },
             )?;
 
-            // Read input by keystroke so that we can suppress ascii control characters
-            if !term.features().is_attended() {
```
tbergerd, line 317: note : removed this because the check is already performed at the beginning of the function (and returns an error instead which makes more sense)

src/prompts/input.rs
```diff
@@ -281,20 +280,47 @@ where
...
+            .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "Quit not allowed in this case"))?)
+    }
+
+    /// Like [`interact_text_opt`](Self::interact_text_opt) but allows a specific terminal to be set.
+    #[inline]
+    pub fn interact_text_on_opt(self, term: &Term) -> Result<Option<T>> {
+        self._interact_text_on(term, true)
+    }
+    
+    fn _interact_text_on(mut self, term: &Term, allow_quit: bool) -> Result<Option<T>> {
```
Gordon01, line 323: You should follow a builder pattern here, so the signature of this function shouldn't change. 
Instead, add a new private field to the `Input`, and a corresponding setter for it.
Access it here via `self`.

src/prompts/input.rs
```diff
@@ -332,6 +353,13 @@ where
 
             loop {
                 match term.read_key()? {
+                    Key::Escape if allow_quit => {
+                        term.clear_line()?;
+                        render.clear()?;
+                        term.flush()?;
+                        term.show_cursor()?;
+                        return Ok(None);
```
Gordon01, line 361: Let's instead add a new variant for the `Error` like `UserInterrupted` and return it here.

src/prompts/input.rs
```diff
@@ -281,20 +280,47 @@ where
 impl<T> Input<'_, T>
 where
     T: Clone + ToString + FromStr,
-    <T as FromStr>::Err: Debug + ToString,
+    <T as FromStr>::Err: ToString,
```
Gordon01, line 283: Why are you removing this bound?

Gordon01 (changes requested):
Thanks for the PR, Please restructure the code a little and it's good to go.

Gordon01:
Oh, I see why you took this approach. Other prompt implementations are like this.
Personally, I think that the builder pattern is cleaner. Single-level `Result` is easier to handle in the user's code and `Error` variants are easily matchable and directly compatible with crates like `anyhow`.

@pksunkara what do you think? Should "cancel" implementations be identical in all prompts?

pksunkara:
> @pksunkara what do you think? Should "cancel" implementations be identical in all prompts?

Yes, but IIRC there were some complications on the `Input` prompt, and I need to review this PR closely this weekend.

src/prompts/input.rs
tbergerd: the bound was only required by the call to unwrap (in the dead code I removed in the same commit : 05f6a3b127d16135e67d535c3eef6d2b1c367d32)
I figured I might as well remove it since it seems unnecessary (interact and interact_on didn't have this constraint).

tbergerd:
Thank you for the review.

Yes I basically copied what was done for other prompts.
If not for consistency, I agree with your points : builder methods + an extra error variant and a single 'interact' would be better (for the impl complexity, the API surface and the calling code)

One thing to note is this wouldn't work with interact and interact_on (as is currently the case with 'completion_with' and 'history_with') though. I'm not quite sure what the use case for those methods is actually (unicode input probably ?)

If it were up to me I'd say consistency should prevail, I'll wait for both your feedbacks.

src/prompts/input.rs
Gordon01: Then it's definitely worth fixing. I suggest making a separate PR for this and removal of the redundant check.

PR #279, opened 2023-09-11
Add '1' and '0' as allowed keys for `Confirm`

Gordon01:
Fixes: https://github.com/console-rs/dialoguer/issues/256

In an effort to foster a more inclusive and accessible user interface, this PR introduces '1' and '0' as alternative input options alongside 'y' and 'n' in the text confirmation prompt.

### DIFF
--- a/src/theme/colorful.rs
+++ b/src/theme/colorful.rs
@@ -153,20 +153,20 @@ impl Theme for ColorfulTheme {
             None => write!(
                 f,
                 "{} {}",
-                self.hint_style.apply_to("(y/n)"),
+                self.hint_style.apply_to("(y/n or 1/0)"),
                 &self.prompt_suffix
             ),
             Some(true) => write!(
                 f,
                 "{} {} {}",
-                self.hint_style.apply_to("(y/n)"),
+                self.hint_style.apply_to("(y/n or 1/0)"),
                 &self.prompt_suffix,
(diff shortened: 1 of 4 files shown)


src/theme/colorful.rs
```diff
@@ -153,20 +153,20 @@ impl Theme for ColorfulTheme {
...
-                self.hint_style.apply_to("(y/n)"),
+                self.hint_style.apply_to("(y/n or 1/0)"),
                 &self.prompt_suffix,
                 self.defaults_style.apply_to("yes")
             ),
             Some(false) => write!(
                 f,
                 "{} {} {}",
-                self.hint_style.apply_to("(y/n)"),
+                self.hint_style.apply_to("(y/n or 1/0)"),
```
pksunkara, line 169: Can we not have the hints in prompts and leave them as y/n?

Gordon01: If you don't want to change the hint, I can make a logic to change it to 1/0 when the non-ascii character is read.
Otherwise, users will probably find it hard to discover alternative keys.

pksunkara: The issue with showing things after a key is pressed is that in some scenarios, the prompt is considered answered as soon as a key is pressed. So, I am wondering if it's worth showing it at all.

Maybe for those scenarios where it waits, we can change it to show 1/0 only if one of them is pressed?

PR #282, opened 2023-09-21
fix #281 : Select : prompt is no longer displayed

### LINKED ISSUE #281: Select : prompt is no longer displayed
This is a regression at head I noticed.
cargo run --example select

Behavior :
The four options get displayed without the prompt
Expected :
the "Pick your flavor" prompt is correctly displayed (works fine in dialoguer 0.10.4)

Specifically the reg was introduced by fix #267
I'm under the impression  Paging::update is only meant to be called after Paging::render_prompt
calling it before makes it so self.activity_transition is set to false (iff there is only one page. notice we do see the prompt correctly on the third part of the 'select' example because it has 2 pages)
here I think we only want to update self.current_page (to fix the issue mentioned in #267)

I'll try and submit a fix but I'm opening the issue in case what I propose is not what we want

tbergerd:
Hello,

I extracted the recomputation of the page into a separate method to fix the default selection not being on the first page without hiding the prompt.

### DIFF
--- a/src/paging.rs
+++ b/src/paging.rs
@@ -50,7 +50,6 @@ impl<'a> Paging<'a> {
             self.current_page = cursor_pos / self.capacity;
         }
     }
-    
 
     /// Updates all internal based on the current terminal size and cursor position
     pub fn update(&mut self, cursor_pos: usize) -> Result {
--- a/src/paging.rs
+++ b/src/paging.rs
@@ -42,6 +42,16 @@ impl<'a> Paging<'a> {
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/paging.rs
```diff
@@ -42,6 +42,16 @@ impl<'a> Paging<'a> {
         }
     }
 
+    pub fn update_page(&mut self, cursor_pos: usize) {
```
tbergerd, line 45: Note : I marked it pub for consistency, otherwise I'd have made it pub(crate). Doesn't matter here because module is private.

pksunkara (approved):

PR #288, opened 2023-10-05
Adding alert as new message type

sluedecke:
In case of an unexpected events, I would love to display a warning to the user, which simply needs to be "confirmed" for the programm to move on.

The workaround I use is a selection with "OK" as single option, but this is a bit hacky.

I kindly request to add alert as a new interaction option to dialoguer. One can use the pull request, but I probably messed the return types up ...

obsoletes https://github.com/console-rs/dialoguer/issues/287

### DIFF
--- a/examples/alert.rs
+++ b/examples/alert.rs
@@ -1,0 +1,16 @@
+use dialoguer::{theme::ColorfulTheme, Alert};
+
+fn main() {
+    let _ = Alert::with_theme(&ColorfulTheme::default())
+        .with_prompt("Something went wrong!  Press enter to continue.")
+        .interact();
+
+    let _ = Alert::with_theme(&ColorfulTheme::default())
+        .with_alert_text("This is an alert, press enter to continue.")
+        .interact();
+
+    let _ = Alert::with_theme(&ColorfulTheme::default())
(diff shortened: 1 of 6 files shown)


pksunkara:
Why can't confirm be used?

sluedecke:
I tried, but it appeared to me that all ways to use Confirm present the user with some kind of choice.  There is always the `y/n` rendered and I am looking for an alert which can only be acknowleged.

This can be done with Confirm and setting a default, but having the 'y/n' suggests that there is a choice if there actually is none.

pksunkara (changes requested):
Instead of adding a new prompt, can you please modify the `confirm` prompt with an extra option?

sluedecke:
Sure.  But I wonder if that would introduce a lot of "if-clauses" into the code making it more complex than needed.  Right now I think a new prompt has these advantages:

* Clearer API - it is focussed on the use case alert only.  How to interpret the options `report`, `default`, `show_default` and `wait_for_newline` for an alert?
* Clearer code - adding it to confirm would make the routines `interact_on`, `theme/mod.rs::format_confirm_prompt` and `theme/colorrful.rs::format_confirm_prompt` more complex.

Probably it is just a mattter of coding style, what do you think?

PR #299, opened 2024-01-18
Accept items by iterator instead of slice

jacobtread:
## Description

Replaces the handling for adding multiple items to the `Select`, `FuzzySelect`, `MultiSelect`, and `Sort` they currently take a slice containing any number of values that can be converted into `String` through `ToString`, this change replaces taking a slice with instead taking anything that implement `IntoIterator` where the item is `ToString`.

This allows the items function to take any collection of values rather than only allowing slices, thus making something like the following possible: 
```rust

struct MyItem {
  name: String,
  // .. other struct items
}

let collection: Vec<MyItem> = vec![  /* Some list of items */ ];
let item_names = collection
  .iter()
  .map(|item| item.name.as_str());

Select::new()
  .items(item_names);
```

## Changes

- Make `items`, `items_checked` (and similar functions) accept an iterator instead of a slice
- Add tests for the new iterator items function

### DIFF
--- a/src/prompts/fuzzy_select.rs
+++ b/src/prompts/fuzzy_select.rs
@@ -81,10 +81,14 @@ impl FuzzySelect<'_> {
     }
 
     /// Adds multiple items to the fuzzy selector.
-    pub fn items<T: ToString>(mut self, items: &[T]) -> Self {
-        for item in items {
-            self.items.push(item.to_string());
-        }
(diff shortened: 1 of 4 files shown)


pksunkara (approved):

Gordon01:
I believe https://github.com/console-rs/dialoguer/pull/223 may now be closed

PR #315, opened 2024-07-04
Added files and folder selection prompt

Hahihula:
Needed option to let the user browse for files or folders.

Maybe it could be useful to other people too.

here is asciicast demo:

[asciicast](https://asciinema.org/a/MaETYSQiMFSqf3V0a3IzUsPB9)


jacek-kurlit:
That would be so useful!

PR #323, opened 2025-02-22
Confirmation prompt terminal width

omertuc:
The confirmation prompt should take into account the terminal's width in relation to the prompt width when clearing the last line(s).

See recording showing before and after: https://asciinema.org/a/wW9IyNYeOT22m9iIH5OPmTki9

### DIFF
--- a/src/prompts/confirm.rs
+++ b/src/prompts/confirm.rs
@@ -163,7 +163,7 @@ impl Confirm<'_> {
             None
         };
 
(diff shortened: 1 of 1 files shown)


omertuc:
Just noticed https://github.com/console-rs/dialoguer/pull/175

cgwalters (commented):
I don't know much about this codebase, but this change looks superficially plausible to me at least!

PR #318, opened 2025-01-06
Upgrade thiserror to v2

paolobarbolini:
https://github.com/dtolnay/thiserror/releases/tag/2.0.0

PR #327, opened 2025-06-22
refactor: replace `thiserror` with a manual impl

CosmicHorrorDev:
Supercedes #318 as there's no need to update `thiserror` if it's no longer used :grin:

It was easy enough to match the functionality provided by `thiserror` with a manual implementation instead

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,6 @@ tempfile = { version = "3", optional = true }
 zeroize = { version = "1.1.1", optional = true }
 fuzzy-matcher = { version = "0.3.7", optional = true }
 shell-words = "1.1.0"
-thiserror = "1.0.40"
 
 [[example]]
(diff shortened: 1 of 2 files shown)


tisonkun (approved):
cc @pksunkara @mitsuhiko

tisonkun:
cc @djc

djc (approved):

djc:
Until I have crates.io privileges for this crate, there is not much point in merging.

tisonkun:
@djc Thanks for your information and really thanks for your maintenance on these crates ..

So we still need to have some time from @mitsuhiko or @pksunkara to add a new crate owner? Do you have their contact to notify them this case?

djc:
I asked @mitsuhiko on July 25th -- I just pinged him with a reminder.

PR #345, opened 2026-04-05
feat: add show_hint option to Confirm prompt

opmr0:
Adds a `show_hint` builder method to `Confirm` that allows hiding the `[y/n]` hint and prompt suffix. This is useful for simple "press enter to continue" use cases where the hint is noise.

**Usage:**
```rust
Confirm::new()
    .with_prompt("Press enter to continue")
    .show_hint(false)
    .default(true)
    .interact()
    .unwrap();
```

**Changes:**

Added `show_hint` builder method to Confirm that allows hiding the [y/n] hint. Implemented via a new `format_confirm_prompt_no_hint` trait method with a default implementation

--- 
Resolves #344

### DIFF
--- a/src/prompts/confirm.rs
+++ b/src/prompts/confirm.rs
@@ -35,6 +35,7 @@ pub struct Confirm<'a> {
     show_default: bool,
     wait_for_newline: bool,
     theme: &'a dyn Theme,
+    show_hint: bool,
 }
 
 impl Default for Confirm<'static> {
@@ -94,6 +95,14 @@ impl Confirm<'_> {
     /// The default is to append the default value to the prompt to tell the user.
     pub fn show_default(mut self, val: bool) -> Self {
         self.show_default = val;
+        self
+    }
+
+    /// Disables or enables the hint display (e.g. `[y/n]`).
+    ///
+    /// The default is to show the hint.
+    pub fn show_hint(mut self, val: bool) -> Self {
+        self.show_hint = val;
         self
     }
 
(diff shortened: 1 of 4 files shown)


src/prompts/confirm.rs
```diff
@@ -163,7 +172,12 @@ impl Confirm<'_> {
             None
         };
 
-        render.confirm_prompt(&self.prompt, default_if_show)?;
+        term.clear_line()?;
+        if self.show_hint {
+            render.confirm_prompt(&self.prompt, default_if_show)?;
+        } else {
+            term.write_str(&format!("{} ", &self.prompt))?;
+        }
```
djc: Why add the `clear_line()`?

Why not use `render.confirm_prompt()` for the `!self.show_hint` case? Looks like this takes care of a bunch of other stuff (like theming) that we wouldn't want to necessarily skip.

djc (commented):
I've resolved the pre-existing CI issues, please rebase and squash your changes into a single commit.

src/prompts/confirm.rs
opmr0: Would you prefer passing `show_hint` through to `format_confirm_prompt` in the theme trait, or is there another approach you have in mind?

djc: > Would you prefer passing `show_hint` through to `format_confirm_prompt` in the theme trait, or is there another approach you have in mind?

I don't have an opinion right now -- happy to review your solution.

opmr0: Would adding a new trait method `format_confirm_prompt_no_hint` with a default implementation work? That way existing theme implementors don't break

djc: Yes, that sounds plausible.

djc (approved):

djc:
Thanks!

opmr0:
Glad to contribute with you 🤍

PR #347, opened 2026-05-22
Fix line length in Select when line has ASCII colors

LandStander27:
I was trying to use `Select`, but because I have ASCII color codes in my items, it was calculating the line length incorrectly, causing problems with rendering. For example, pressing `Up` would move the entire selection prompt up as well, deleting the line above the prompt.

So, I made a function `len_visible()` on `&str` that calculates the length, excluding the color codes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 target
+/rustfmt.toml
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,1 @@
 target
-/rustfmt.toml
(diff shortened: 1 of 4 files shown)


djc:
There's API for this in the console crate which dialoguer already depends on -- would be better to reuse that.

LandStander27:
Sorry, didn't see that.

djc:
Can you do a separate PR for resolving the clippy lints, please?

PR #353, opened 2026-07-02
Fix FuzzySelect returning wrong index for duplicate items

### LINKED ISSUE #286: FuzzySelect returns wrong index if items has duplicate entries
If a slice with duplicate values is passed to the `items()` method on a new `FuzzySelect`, the index of the first item is returned if the user selects the second item.

```rust
use dialoguer::{console::Term, FuzzySelect};

fn main() {
    let items = ["item0", "item1", "item1", "item3", "item4"];

    let selection = FuzzySelect::new()
        .items(&items)
        .interact_on_opt(&Term::stderr())
        .unwrap();

    if let Some(index) = selection {
        println!("Selection: {index}");
    }
}
```

Selecting the first item returns the correct index.
```
|
  item0
> item1
  item1
  item3
  item4

Selection: 1
```

Selecting the second item returns the incorrect index.
```
|
  item0
  item1
> item1
  item3
  item4

Selection: 1
```

apoorva-01:
Picking a duplicate in `FuzzySelect` gave back the wrong index. It recovered the selection by finding the first item equal to the chosen string, so any dup resolved to its first occurrence. Selecting the second `item1` in `["item0", "item1", "item1", "item3", "item4"]` returned `1` instead of `2`.

Fix threads each item's original index through the filtered list with `enumerate` and returns that, instead of looking the string back up.

I pulled the filter/rank step into a small private `filter_and_rank` so it's unit-testable (the selection lives in the interactive `read_key` loop with no keypress harness, so the test covers the index resolution, which is the actual bug). Could also just inline the `enumerate` fix without the helper if you'd rather, but then there's nothing to test against.

closes #286


src/prompts/fuzzy_select.rs
```diff
@@ -238,18 +238,9 @@ impl FuzzySelect<'_> {
...
             render.fuzzy_select_prompt(self.prompt.as_str(), &search_term, byte_indices[cursor])?;
 
-            // Maps all items to a tuple of item and its match score.
-            let mut filtered_list = self
-                .items
-                .iter()
-                .map(|item| (item, matcher.fuzzy_match(item, &search_term)))
-                .filter_map(|(item, score)| score.map(|s| (item, s)))
-                .collect::<Vec<_>>();
+            let filtered_list = filter_and_rank(&self.items, &matcher, &search_term);
```
djc, line 241: Please split this in two commits: one that extracts a `filter_and_rank()` function, and another that makes changes to the filtering/ranking.

apoorva-01: Done, split into the extract and the fix.

djc:
If they're duplicates, why does it matter what index we resolve to?

Should we include the index in the comparison so that duplicate items always appear in the same order?

src/prompts/fuzzy_select.rs
```diff
@@ -398,6 +388,26 @@ impl<'a> FuzzySelect<'a> {
...
+        .filter_map(|item| {
+            matcher
+                .fuzzy_match(item, search_term)
+                .map(|score| (item, score))
+        })
+        .collect::<Vec<_>>();
+
+    filtered.sort_unstable_by(|(_, s1), (_, s2)| s2.cmp(s1));
+
+    filtered.into_iter().map(|(item, _)| item).collect()
```
djc, line 408: If I'm not mistaken, we can drop the `collect()` call here (in the first commit) and yield an `impl Iterator<Item = &'a String>`, since the caller consumes an iterator anyway.

PR #355, opened 2026-07-07
feat: Add support for masked password prompt

TomPlanche:
## Summary

Adds an optional mask character to the `Password` prompt. By default the input stays fully hidden (nothing echoed) exactly as before; when a mask is set, one mask character is echoed per typed character.

Resolves #340.

```rust
let password = Password::new()
    .with_prompt("Enter password")
    .with_mask('*') // if not set, input stays fully hidden
    .interact()?;
```

## Changes

- New `with_mask(char)` builder method on `Password`, backed by a `mask: Option<char>` field (`None` by default).
- When a mask is set, input is read key-by-key via `Term::read_key()` and a mask character is echoed for each printable character. `Backspace` erases the last character; `Enter` submits. This path mirrors `read_secure_line`'s trailing-newline behaviour so both code paths behave uniformly.
- When no mask is set, behaviour is unchanged (`read_secure_line`).
- Extended `examples/password.rs` to demonstrate the masked prompt.

## Notes

- **Non-breaking**: purely additive. The new `Option<char>` field defaults to `None`, preserving the current fully-hidden behaviour.
- **Zeroize**: `read_masked_line` returns a `String` and the caller wraps it in `Zeroizing` in `interact_on`, identical to the existing `read_secure_line` path, so the existing memory-scrubbing guarantee is unchanged.


examples/password.rs
```diff
@@ -18,4 +18,15 @@ fn main() {
         "Your password is {} characters long",
         password.chars().count()
     );
+
+    let masked = Password::with_theme(&ColorfulTheme::default())
```
djc: Instead of adding another thing here, add a commented out `with_mask()` call with a one-line comment to the existing code?

src/prompts/password.rs
```diff
@@ -145,6 +175,35 @@ impl Password<'_> {
             }
         }
     }
+
+    /// Reads a line while echoing a mask character for each typed character.
+    ///
+    /// This mirrors the trailing newline behaviour of
+    /// [`Term::read_secure_line`] so the caller can treat both paths uniformly.
+    fn read_masked_line(term: &Term, mask: char) -> Result<String> {
```
djc: I'd prefer to inline this into the single caller.

src/prompts/password.rs
```diff
@@ -136,7 +163,10 @@ impl Password<'_> {
             render.password_prompt(prompt)?;
             render.term().flush()?;
 
-            let input = render.term().read_secure_line()?;
+            let input = match self.mask {
+                Some(mask) => Self::read_masked_line(render.term(), mask)?,
```
djc: When inlining `read_masked_line()`, suggest rewriting this with let-else so we do an early return for non-masked case.

src/prompts/password.rs
```diff
@@ -136,12 +163,43 @@ impl Password<'_> {
...
+            };
+
+            let term = render.term();
+            let mut password = String::new();
+
+            loop {
+                match term.read_key()? {
+                    Key::Char(chr) if !chr.is_ascii_control() => {
+                        password.push(chr);
+                        term.write_str(&mask.to_string())?;
```
djc: Sorry, missed this in my first round: the allocation here is a bit ugly and should be unnecessary. I think `write_fmt("{mask}")` could work?

djc (approved):
Looks okay modulo one nit. Please squash your commits.

src/prompts/password.rs
TomPlanche: `write_fmt` takes `&mut self` (it's from `io::Write`), but `term` here is a shared `&Term`

djc: Well, `write_str()` is hidden as well, so we probably shouldn't be using it.

TomPlanche: Do you prefer: `term.write_fmt(format_args!("{}", mask))?;` or `write!(term, "{mask}")?;`, they both expands to the same I think.

djc: The former wouldn't work because of the shared reference, right? Am I missing something?

TomPlanche: It works now

djc: I think the other thing that could work is something like this:

```rust
let mut buf = [0; 4];
term.write_all(mask.encode_utf8(&mut buf));
```

TomPlanche: `term.write_str(mask.encode_utf8(&mut [0; 4]))?;` do not require `term` to be mutable, perfect

djc:
Thanks!

PR #363, opened 2026-09-15
Use post completion text when Input default is accepted

### LINKED ISSUE #309: `Input` post completion text does not appear if the user uses the default
I have a `Input` that looks like this:
```rs
let result: String = Input::new()
        .with_prompt("My Prompt")
        .with_post_completion_text("Completed Prompt")
        .default("My Default".to_string())
        .interact_text()
        .unwrap();
```
If the user types in their own input, everything works as intended. However if the user just presses <kbd>Enter</kbd> to use the default string, then the post completion text does not appear, instead the prompt appears.

ethanstoner:
Fixes #309

When the user pressed Enter to accept the default in `Input::interact_text`, the report line always rendered `self.prompt`; only the typed-input path checked `post_completion_text`. Both paths now go through a small private helper, `completion_prompt()`, so they can't drift apart again.

A second commit fixes the clippy lints that newer stable Rust (1.98) reports in untouched code (`useless_borrows_in_formatting` in the theme code, `needless_return_with_question_mark` in `select`/`multi_select`/`sort`), which were failing the Linting job on `main` too. It's mechanical (`cargo clippy --fix`) with no behavior change; happy to drop it or split it into its own PR if you prefer.

## Testing

- Added the unit test `test_completion_prompt`. It covers the helper, because `interact_text_on` requires a real terminal.
- Reproduced with the example from the issue, driven through a PTY:
  - Before, pressing Enter printed `My Prompt: My Default`.
  - After, it prints `Completed Prompt: My Default`, which matches the typed path (`Completed Prompt: abc`).
- `cargo fmt --check`, `cargo clippy --locked --all-targets --all-features -- -D warnings` (rustc 1.98.1), `cargo test --locked --all-features` and `cargo doc` with `-Dwarnings` all pass.

`interact()` / `interact_on` doesn't use `post_completion_text` on either path. I left that out to keep this PR focused, but it's a two-line follow-up with the same helper if you want it.

This change was made with the help of Claude Code; I reviewed and tested it.


djc (approved):
Looks okay.

I had addressed the clippy failures already in

- #362 

but had failed to merge it. Please rebase on top of current `main`.

ethanstoner:
Rebased on current `main`. Since #362 landed the Clippy fixes, I dropped my "Fix clippy lints reported by Rust 1.98" commit as redundant — the branch is now the single fix commit.

Verified locally on top of `af5db9d`: `cargo fmt --check` clean, `cargo clippy --all-features --all-targets -- -D warnings` clean, and `cargo test --all-features` passes (15 unit + 26 doc tests).

